PR #17, opened 2021-02-20
fix: species cleared on approval

### LINKED ISSUE #15: Species is only recorded on first approval
When verifying captures, the species stays in the species box after _Submit_ is clicked, suggesting that it is retained for subsequent approvals.
In fact it is being cleared in the model (and not reflected in the UI) so that the species only applies to the first approval.

See [this video of the bug in action](https://youtu.be/dBkZpkrKW_k).

nmcharlton:
Resolves #15 

This PR contains a few related fixes and improvements, as well as having _Prettier_ run on all modified files.

## Clearing Approval Selection
By default, approval side panel selections are now all reset when a user presses _Submit_
A new _Remember selection_ checkbox gives the user the option to prevent the reset for the session

Screenshot 2021-02-20 at 22 31 53

## Submit Button
The _Submit_ button is now disabled if no captures are selected

## Captures per Page
The maximum captures per page has been doubled to 192 on @sebastiangaertner's request.

## File/Model Naming
The misnamed `verity` model has been renamed `verify`


rematchTutorial.md
```diff
@@ -337,21 +338,21 @@ const model = {
 		async loadMoreTreeImages(payload, state){
 			//{{{
 			log.debug('to load images')
-			const verityState		= state.verity
-			if (verityState.isLoading || !verityState.moreTreeImagesAvailable){
+			const verifyState		= state.verify
```
dadiorchen, line 341: Yes, I think you can guess who is the bad guy with poor English did this.  😉 
By the way, is it easy to use IDE like vscode to refactor the renaming? Is any manual work needed?

dadiorchen (approved):

rematchTutorial.md
nmcharlton: I'm sure there was a better way, but I just did a find and replace on instances of `verity`.

PR #20, opened 2021-02-22
feat: add optimized images

### LINKED ISSUE #18: optimizing performance through resizing images delivering them through CDN
Currently an image is displayed at around 480*420 px in a three column view before it jumps to the 4 column view. The actual image is delivered and loaded at around 768*1024px. The image could easily resized via CDN to about 1/3rd of its size in kB. Since most admin panel operators are in countries with slow internet this is a critical improvement and a low hanging fruit. Currently Freetown has to verify 50.000 trees and their internet is slow creating a bottle neck.

Possible services if not part of AWS or DO are https://statically.io/ 

Below two images ... one is whats loaded by the browser the other one is the size displayed. 

Screenshot 2020-11-20 at 06 10 54

luacmartins:
Closes #18

### DIFF
--- a/src/components/OptimizedImage.js
+++ b/src/components/OptimizedImage.js
@@ -15,29 +15,34 @@ export default function OptimizedImage(props) {
   if (!src) return <></>;
 
   const cdnPath = 'https://cdn.statically.io/img';
-  const domain = src.match(/\/\/(.*)\//)[1];
-  const imagePath = src.match(/.com\/(.*)/)[1];
-  const params =
-    `f=auto,w=${width}` +
-    (height ? `,h=${height}` : '') +
-    (quality ? `,q=${quality}` : '');
+  const matches = src.match(/\/\/(.*)\/(.*)/);
 
-  const cdnUrl = `${cdnPath}/${domain}/${params}/${imagePath}`;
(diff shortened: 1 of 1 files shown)


src/components/OptimizedImage.js
```diff
@@ -0,0 +1,56 @@
...
+    imageSizes = [400, 300, 250, 200],
+    fixed,
+    ...rest
+  } = props;
+
+  if (!src) return <></>;
+
+  const cdnPath = 'https://cdn.statically.io/img';
+  const domain = src.match(/\/\/(.*)\//)[1];
+  const imagePath = src.match(/.com\/(.*)/)[1];
```
nmcharlton, line 19: I think this is a reasonable assumption, but it would also be good to fall back to the original image if we can't match the image path or domain.

You could combine and generalise these two checks to:
```js
const matches = src.match(/\/\/(.*)\/(.*)/);
if (matches.length > 2) {
  const domain = matches[1];
  const imagePath = matches[2];
  ...
}
```

luacmartins: Thanks for the great feedback @nmcharlton! I made the necessary changes.

Issue #26, opened 2021-02-24
Filter by Organization ID


sebastiangaertner:
We need to be able to filter by organization ID including the org. hierarchy.

nmcharlton:
@sebastiangaertner Are you referring to the capture filter in the Verify tool?

The basic version of the organization filter was implemented in December under [PR #482](https://github.com/Greenstand/treetracker-admin-api/pull/482) (see dev deployment – https://d2m5zl0de9jrzz.cloudfront.net) buit has not made it into the test environment yet.

Currently, all orgs are listed, and selecting a parent org in the filter includes all child org captures:

![Screenshot 2021-02-24 at 21 28 27](https://user-images.githubusercontent.com/5558838/109068052-414cb080-76e7-11eb-887b-e1ed819b62bf.png)

We have a design for a more complex implementation that involves preset exclude filter configurations:

![OrgFilter (1)](https://user-images.githubusercontent.com/5558838/109069511-385cde80-76e9-11eb-8ad6-08404210ee99.png)

gwynndp:
I'm looking into this!

nmcharlton:
We need a way to distinguish planting organizations from other organizations in the entity/stakeholder table.
Only the former should appear in the organization filter.

nmcharlton:
This will be resolved by showing only entities with `type='O'` (see issue #73).

nmcharlton:
Resolved in #87

PR #38, opened 2021-03-06
feat: add reset button to tree filter

gwynndp:
This pull is just exploration, open to feedback if I'm going the wrong direction with changes.

### DIFF
--- a/src/components/FilterTop.js
+++ b/src/components/FilterTop.js
@@ -68,13 +68,11 @@ function Filter(props) {
   const { classes, filter } = props;
   const dateStartDefault = null;
   const dateEndDefault = null;
-  const [treeId, setTreeId] = useState(filter.treeId);
-  const [planterId, setPlanterId] = useState(filter.planterId);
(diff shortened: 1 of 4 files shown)


nmcharlton:
Hi @gwynndp, the issue you've mentioned looks unrelated to this change.
Is this addressing an open issue, or just a general improvement?

gwynndp:
I was exploring how the filter worked related to this issue: https://github.com/Greenstand/treetracker-admin-client/issues/26
And added a reset button because it was useful.  I'm not clear what the spec is for the the filter so this is just a draft for the purpose of checking in.

nmcharlton (changes requested):
I'm happy with the reset button implementation, but we should remove the client-side filtering and fix it properly on the server side.

Greenstand/treetracker-admin-api#542 captures the Organization filter bug.
If there are any other problems with what the legacy API returns when a filter is applied, we can raise issues for those against the API project.

nmcharlton (approved):
Happy days 😎

PR #42, opened 2021-03-11
Feat: Add country

### LINKED ISSUE #13: Show country of registration in Planter details
We should include the planter's country of registration in the Planter Details dialog.

Screenshot 2020-10-02 at 16 10 05

luacmartins:
Closes #13

### DIFF
--- a/src/components/PlanterDetail.js
+++ b/src/components/PlanterDetail.js
@@ -83,7 +83,10 @@ const PlanterDetail = (props) => {
           setPlanterRegistrations(null);
           api.getPlanterRegistrations(planterId).then((registrations) => {
             if (registrations && registrations.length) {
-              setPlanterRegistrations(registrations);
+              const sortedRegistrations = registrations.sort((a, b) =>
+                a.created_at > b.created_at ? 1 : -1,
+              );
+              setPlanterRegistrations(sortedRegistrations);
             }
           });
         }
@@ -91,12 +94,7 @@ const PlanterDetail = (props) => {
     }
     loadPlanterDetail();
     // eslint-disable-next-line
-  }, [
-    planterId,
-    planterRegistrations,
-    props.plantersState.planters,
-    props.plantersDispatch,
-  ]);
+  }, [planterId, props.plantersState.planters, props.plantersDispatch]);
 
   function handleEditClick() {
     setEditDialogOpen(true);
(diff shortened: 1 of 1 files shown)


src/components/PlanterDetail.js
```diff
@@ -78,18 +77,17 @@ const PlanterDetail = (props) => {
 
         if (
           !planterRegistration ||
-          planterRegistration.planterId !== planterId
+          planterRegistration[0].planter_id !== planterId
```
nmcharlton, line 80: This will throw an exception if the previous query returned no `planter_registrations`.
We need to check `planterRegistration.length` first.

nmcharlton, line 80: Can we change the name of the variable to `planterRegistrations` to indicate that it's now an array?

src/components/PlanterDetail.js
```diff
@@ -195,11 +193,29 @@ const PlanterDetail = (props) => {
...
+              </Typography>
+            </Grid>
+            <Divider />
             <Grid container direction="column" className={classes.box}>
               <Typography variant="subtitle1">Registered</Typography>
               <Typography variant="body1">
                 {(planterRegistration &&
-                  getDateTimeStringLocale(planterRegistration.createdAt)) ||
+                  getDateTimeStringLocale(
+                    planterRegistration[planterRegistration.length - 1]
```
nmcharlton, line 216: We aren't guaranteed that the last registration in the array will be the earliest.
We could either order by registration date in the API query or perform a sort when the results are returned.

nmcharlton (changes requested):
Just some minor suggestions

src/components/PlanterDetail.js
luacmartins: Oops, thanks for picking this up!

src/components/PlanterDetail.js
luacmartins: You're right there. I will implement a sort on the frontend.

src/components/PlanterDetail.js
```diff
@@ -195,11 +194,31 @@ const PlanterDetail = (props) => {
...
+            <Divider />
             <Grid container direction="column" className={classes.box}>
               <Typography variant="subtitle1">Registered</Typography>
               <Typography variant="body1">
-                {(planterRegistration &&
-                  getDateTimeStringLocale(planterRegistration.createdAt)) ||
+                {(planterRegistrations &&
+                  planterRegistrations.length > 0 &&
+                  getDateTimeStringLocale(
+                    planterRegistrations.sort((a, b) =>
```
nmcharlton, line 218: It would be simpler to do the sort when the registrations are loaded, rather than on each render.

src/components/PlanterDetail.js
```diff
@@ -195,11 +194,31 @@ const PlanterDetail = (props) => {
...
+                      a.created_at > b.created_at ? -1 : 1,
```
nmcharlton, line 219: This gives us the _latest_ registration date, whereas we want the _earliest_.

nmcharlton (changes requested):

nmcharlton (approved):
Perfect, thanks a lot for your efforts @luacmartins

PR #62, opened 2021-04-08
fix: disable service worker

### LINKED ISSUE #60: Problems with web app caching in production
We are seeing issues in production where previously cached versions of the Admin Panel are being loaded by the browser.
This is problematic because we have recently changed the API url and the cached version of the Admin Panel points at the old url.

Refreshing the Admin Panel should cause the latest version to replace the cached version but we've had reports of people having to manually clear their browser cache/storage to get the new version.

This may be related to a 404 error seen on first load when trying to retrieve `service-worker.js` (this file has not yet been created for the Admin Panel).

nmcharlton:
May resolve #60 (to be tested in dev and test environments).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,6 @@ import { init } from '@rematch/core';
 import * as models from './models';
 import './index.css';
 import * as log from 'loglevel';
-import registerServiceWorker from 'registerServiceWorker';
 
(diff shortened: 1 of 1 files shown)


nmcharlton:
Pushing PR through to test on dev deployment

PR #67, opened 2021-04-14
feat: add planter map link to capture thumbnail in Verify tool

### LINKED ISSUE #12: Link to Planter id (currently still in webmap as userid) from each image
See attached proposed design. Next to the location link (which does not work since the individual tree link https://github.com/Greenstand/treetracker-web-map/issues/178 ) we need a link to the userid such as https://treetracker.org/?userid=1868
link-to-planter_id

### LINKED ISSUE #49: Improve quick links on verification tool
This issue from Seb was found in the web map repo. It belongs here

See below the two images. Second icon should open the users map to quickly see the overview. (such as https://treetracker.org/?userid=3336)
Changing the order of icons first the pin icon with link to individual_tree, then the user_info_icon and then the user_map_icon
usermap_link
userinfo_link

### LINKED ISSUE #64: Optimized capture thumbnail images are too low resolution on larger screens
The Verify page uses optimized images for its thumbnails, but they are too small on larger screens.
In a 1600px wide window, the thumbnails are 388px wide but the images are optimized to 320px.
This causes blurry images.

With the current breakpoint configuration, the thumbnails go up to 495px wide.

Screenshot 2021-04-09 at 19 30 28

### LINKED ISSUE #65: Fix 'trees to verify' label
The 'trees to verify' label only makes sense when unverified trees are selected in the filter.
It should be changed to simply 'trees', or better still, 'captures' (see #3).

We should also use locale formatting (as seen in the dash stats) to ensure the number is easily readable when large.

Screenshot 2021-04-13 at 22 44 49

nmcharlton:
- Resolves #12 
- Resolves #49
- Resolves #64 
- Resolves #65 

1. Added map link to capture verification card
2. Increased thumbnail width from 320px to 400px
3. Increased tree detail image width from 480px to 640px
3. Changed _trees to verify_ to _captures_ and added locale formatting to number

https://user-images.githubusercontent.com/5558838/114778711-1c7ecc00-9d6d-11eb-9be9-dfce9e73503b.mov

### DIFF
--- a/src/components/TreeImageScrubber.js
+++ b/src/components/TreeImageScrubber.js
@@ -393,7 +393,7 @@ const TreeImageScrubber = (props) => {
             <CardContent className={classes.cardContent}>
               <OptimizedImage
                 src={tree.imageUrl}
-                width={480}
+                width={400}
                 className={classes.cardMedia}
               />
             </CardContent>


dadiorchen:
@nmcharlton I am using "Loom" these days for record, it's super convenient, just a Click on the page :D 

Screen Shot 2021-04-16 at 10 11 38 AM

dadiorchen (approved):

PR #69, opened 2021-04-17
Feature/html title

### LINKED ISSUE #39: Improve HTML title for Treetracker Admin web pages
The current title, **Greenstand - Tree Tracker**, is not very helpful and breaks the Greenstand communications guidance by splitting Treetracker into two words.

For consistency with the webmap, it should read **Treetracker Admin by Greenstand**

The title should also change dynamically with page changes, e.g. **Verify - Treetracker Admin by Greenstand**

VLuisa:
* Resolves #39 

Default html title is now set to **Treetracker Admin by Greenstand** which is also the title for the 'Home' page.
All other pages are dynamically retitled formatted as: `[Page title] - Treetracker Admin by Greenstand`

I used the Effect hook for the dynamic title updates - open to suggestions if there are better methods!

image
image

### DIFF
--- a/src/components/CaptureMatching/CaptureMatchingFrame.js
+++ b/src/components/CaptureMatching/CaptureMatchingFrame.js
@@ -134,7 +134,7 @@ function CaptureMachineFrame() {
   /* to update html document title */
   useEffect(() => {
     document.title = `Capture Matching - ${documentTitle}`;
-  });
+  }, []);
 
   return (
     <>
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -157,7 +157,7 @@ function Home(props) {
   /* to update html document title */
   useEffect(() => {
     document.title = `${documentTitle}`;
-  });
+  }, []);
 
   return (
     <div className={classes.box}>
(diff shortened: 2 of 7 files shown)


VLuisa:
@nmcharlton I also seem to have fallen prey to a lot of reformatting changes - not sure if there's something I should have done to avoid that?

nmcharlton (changes requested):
Thanks @VLuisa, this looks great!
I agree that `useEffect()` is the best way to change the title.

One suggestion is to pass an empty array as the second parameter (`dependencies`) to `useEffect()` to ensure it is only called once per load.
With no explicit dependencies, the callback is called every time the React render loop fires.

Regarding the reformatting, we have Prettier configured for the project to apply code format standards.
Depending on how your environment is set up, formatting will be applied on either save or commit. It's likely that the code you haven't touched had previously evaded formatting.

VLuisa:
@nmcharlton Thanks for the tip, just made the changes

nmcharlton (approved):
Perfect, great work @VLuisa!

PR #76, opened 2021-04-30
Feature/case insensitive name

### LINKED ISSUE #72: Planter name filter should be case-insensitive
Go to live DB and search planter with first name "Deepak" and it does not pick up planter 3571. Where is the first name looked up? In the planter or planter registration ? see video link attached ... https://youtu.be/6K7-pBVatf0

VLuisa:
Resolves #72 
* Using `ilike` operator to match case insensitive regular expressions for `firstName` and `lastName`
* Refactored Planter to use filter object instead of string

image

### DIFF
--- a/src/models/FilterPlanter.test.js
+++ b/src/models/FilterPlanter.test.js
@@ -12,33 +12,14 @@ describe('Filter', () => {
     });
   });
 
-  it('getBackloopString() should be: ', () => {
-    expect(filterPlanter.getBackloopString()).toEqual(
-      expect.stringContaining('&filter[where][personId]=1'),
-    );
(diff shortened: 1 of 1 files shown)


VLuisa:
@nmcharlton Ah I didn't check the tests, and I see now it's looking for the `getBackloopString()` function I removed... Should I update the test?

```
it('getBackloopString() should be: ', () => {
    expect(filterPlanter.getBackloopString()).toEqual(
      expect.stringContaining('&filter[where][personId]=1'),
    );
```

nmcharlton (approved):
Excellent work, thanks @VLuisa!

Issue #58, opened 2021-03-29
Prevent user from creating a new species with the same name as an existing one


nmcharlton:
Users can add species through the Species page or by tagging a capture with a species name they type in (rather than selecting from the dropdown).

Species names should be unique; if the name already exists, the capture should be tagged with that species.

nmcharlton:
The species name matching should also be case-insensitive.

tranquanghuy0801:
@nmcharlton , can I have a look at this issue?

tranquanghuy0801:
![Screen Shot 2021-05-01 at 10 47 18](https://user-images.githubusercontent.com/29462498/116765841-9cde4600-aa6a-11eb-8a19-955ff2499aa7.png)
Hi @nmcharlton, for each capture at this page, is that multiple species can be tagged on the settings, right ?

tranquanghuy0801:
Check my new PR #77

nmcharlton:
@tranquanghuy0801 We can only assign one species to each capture or set of captures.
If a user types in a new species when tagging a capture, they are asked if they want to create a new species:
![Screenshot 2021-05-01 at 14 10 21](https://user-images.githubusercontent.com/5558838/116783513-f8b6c800-aa86-11eb-8f97-da54ee68c57b.png)
If what they've typed matches the name of an existing species (case-insensitive), we should use the existing one instead of showing this alert.

nmcharlton:
Fixed in PR #77

PR #77, opened 2021-05-01
Fixed creating species with the same name

tranquanghuy0801:
Resolves issue #58 

- Add a notification "Species already exist" if user create new species with the same name
- Tagging species in a capture already created new species if they did not exist

### DIFF
--- a/src/components/SpeciesTable.js
+++ b/src/components/SpeciesTable.js
@@ -374,7 +374,7 @@ const EditModal = ({
   data,
 }) => {
   const [error, setError] = React.useState(undefined);
-  const nameSpecies = data.map((species) => species.name);
+  const nameSpecies = data.map((species) => species.name.toLowerCase());
 
   const onNameChange = (e) => {
     setError(undefined);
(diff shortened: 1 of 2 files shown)


src/components/SpeciesTable.js
```diff
@@ -384,14 +391,19 @@ const EditModal = ({
...
   const handleSave = async () => {
-    setIsEdit(false);
-    await editSpecies({
-      id: speciesEdit.id,
-      name: speciesEdit.name,
-      desc: speciesEdit.desc,
-    });
-    loadSpeciesList();
-    setSpeciesEdit(undefined);
+    if (nameSpecies.includes(speciesEdit.name.trim())) {
```
nmcharlton, line 394: Can we make this a case-insensitive comparison?

nmcharlton (changes requested):
This looks pretty good for the species table, just one suggestion about case-insensitive name matching.

We also need to make the `isNewSpecies` check in `src/models/species.js` case-insensitive.

src/components/SpeciesTable.js
tranquanghuy0801: Yes. I have fixed in my latest commit.

nmcharlton (approved):
Awesome work 🌴

Issue #40, opened 2021-03-09
Use OptimizedImage in ImageScroller


nmcharlton:
The `OptimizedImage` class compresses images to thumbnails of a more appropriate size.
The `ImageScroller` class (used by `EditPlanter` to display planter images) does not currently use `OptimizedImage`, so full-size planter images are loaded unnecessarily.

tranquanghuy0801:
Hi @nmcharlton , please check my PR #78

PR #78, opened 2021-05-01
Fixed adding OptimizeImage in ImageScroller of EditPlanter

tranquanghuy0801:
Resolves issue #40


nmcharlton (approved):
Thanks @tranquanghuy0801!

Issue #57, opened 2021-03-29
Planter Manager cannot set planter selfie without Tree Manager role


nmcharlton:
We query trees planted by the selected planter to get a set of selfies.
If a user with a Planter Manager role does not also have the Tree Manager or Admin role, the `trees` query will fail with 401 Unauthorized.

tranquanghuy0801:
Hi @nmcharlton , can I have a look at this issue?

tranquanghuy0801:
![Screen Shot 2021-05-02 at 12 26 08](https://user-images.githubusercontent.com/29462498/116800129-953da080-ab41-11eb-84ab-f34c4ba81a89.png)
Is this issue meaning that Planter Manager without Tree Manager role cannot use the filters in this screen?

nmcharlton:
@tranquanghuy0801 This issue refers to the images in the Edit Planter dialog you worked on in #78.

tranquanghuy0801:
As I have seen from the web app, the images in the Edit Planter are fixed. You cannot edit to change those.

nmcharlton:
@tranquanghuy0801 Sorry for the poor explanation of this issue.

It looks like we don't have many examples in the dev database, but if the planter has taken multiple tree captures, a planter photo will be associated with each capture.
We display each of these photos in the _Edit Planter_ dialog and allow the user to select from them.
In most cases in dev, there is at most one photo available, but in production there will be many.

Here's an example from admin.treetracker.org (with personal details redacted):
![Screenshot 2021-05-02 at 10 19 23](https://user-images.githubusercontent.com/5558838/116808288-e6449900-ab2f-11eb-8bde-891993b50d9e.png)

Try __planter ID 5338__ in dev. That has a couple of photos available.

To test, create a user for yourself with the _Planter Manager_ role but not _Tree Manager_ in the _User Manager_ tool.
Once you've created the user, you'll need to click the key icon to set the password before you can log in as that user.

nmcharlton:
We need to figure out a better way of retrieving the `planterPhotoUrl` for each tree/capture associated with the planter without the need to query `trees` directly from the client.

The best option may be a new API endpoint under `/api/planters` (e.g. `/api/planters/:id/selfies`) that queries the trees controller directly on the server, bypassing the role authentication for the `/api/trees` endpoint.

PR #83, opened 2021-05-07
Fix: Set Planter Manager not update the selfie without Tree Manager Role

tranquanghuy0801:
Resolves issue #57

### DIFF
--- a/src/api/planters.js
+++ b/src/api/planters.js
@@ -85,13 +85,13 @@ export default {
       fields: ['planterPhotoUrl'],
     };
 
-    const treeQuery = `${
+    const planterSelfiesQuery = `${
       process.env.REACT_APP_API_ROOT
     }/api/${getOrganization()}planter/${planterId}/selfies/?filter=${JSON.stringify(
       filter,
     )}`;
 
-    return fetch(treeQuery, {
+    return fetch(planterSelfiesQuery, {
       method: 'GET',
       headers: {
         'content-type': 'application/json',
--- a/src/components/EditPlanter.js
+++ b/src/components/EditPlanter.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useContext } from 'react';
+import React, { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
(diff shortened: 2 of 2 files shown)


tranquanghuy0801:
@nmcharlton  I have also added another PR in Admin API

src/components/EditPlanter.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
```
nmcharlton, line 1: There's no need to make any changes to this file.
These changes can be reverted.

src/components/EditPlanter.js
```diff
@@ -91,7 +94,16 @@ const EditPlanter = (props) => {
   }
 
   function handleSelectPlanterImage(img) {
-    handleChange('imageUrl', img);
+    // Planter Manager with Tree Manager role cannot set planter selfie
```
nmcharlton, line 97: This is not correct.
A planter manager should be able to set this image even if they don't have tree permissions.

src/api/planters.js
```diff
@@ -83,12 +83,13 @@ export default {
       order: 'timeUpdated DESC',
       limit: 100,
       fields: ['planterPhotoUrl'],
-      where: { planterId, planterPhotoUrl: { neq: null } },
     };
 
     const treeQuery = `${
```
nmcharlton, line 88: It doesn't make sense to call this `treeQuery` any more, as it's querying planter selfies.
The client doesn't need to know that the api is using the trees database for this lookup.

nmcharlton (changes requested):

src/components/EditPlanter.js
tranquanghuy0801: Sorry I wrote wrong comment. It should be "Planter Manager without Tree Manager role cannot set planter selfie". 
So what you mean I should hangle logic when clicking "Save" button instead?

nmcharlton: No, I mean that we don’t need to check permissions at all.
The title of the issue is a description of the bug, not the desired functionality.

nmcharlton (approved):
Great work!

Issue #63, opened 2021-04-09
Dashboard stats and trees do not update when logging in as a different user


nmcharlton:
When switching accounts between a non-org and an org user, or between orgs, the data in the Admin Panel should be refreshed to match the new user.

Instead, stale data (dashboard stats, trees in the Verify tool) persists until the page is refreshed.

We should reset all Rematch models when logging out.

https://user-images.githubusercontent.com/5558838/114223303-8d864400-9967-11eb-85eb-d24a4ee35466.mov

tranquanghuy0801:
Hi @nmcharlton, can I have a look at this issue?

PR #84, opened 2021-05-08
Fix: set dashboard stats updating when logging with a different user

tranquanghuy0801:
Resolves issue #63


tranquanghuy0801:
I have updated the part of counting planters

nmcharlton (approved):

Issue #80, opened 2021-05-05
Make Verify and Planters filter buttons more prominent


nmcharlton:
The filter button on the Verify page is hard to find for new users.
It should be made more prominent, perhaps just by adding the word 'Filter' next to the icon.
![Screenshot 2021-05-05 at 18 47 21](https://user-images.githubusercontent.com/5558838/117186126-7144be00-add2-11eb-81fc-9f0d93788a8d.png)

Issue #29, opened 2021-02-24
Pressing the enter key in a filter should trigger Apply


nmcharlton:
In the following filters, the user should be able to trigger an Apply by pressing enter while focussed in one of the filter fields:

- Verify filter
- Trees filter
- Planters filter

We should be able to do this easily using a form submit control.

This goes for any new filters that are created.

![Screenshot 2021-02-24 at 22 26 36](https://user-images.githubusercontent.com/5558838/109074205-5cbbb980-76ef-11eb-872a-e033c910a3a2.png)

PR #85, opened 2021-05-08
Fix: trigger Apply button in Verify tab using enter key and update filter button

tranquanghuy0801:
Resolves issues #29 and #80

### DIFF
--- a/src/components/Planters.js
+++ b/src/components/Planters.js
@@ -229,7 +229,7 @@ const Planters = (props) => {
           <Navbar
             buttons={[
               <Button
-                variant="outlined"
+                variant="text"
                 color="primary"
                 onClick={handleFilterClick}
                 startIcon={<IconFilter />}
--- a/src/components/TreeImageScrubber.js
+++ b/src/components/TreeImageScrubber.js
@@ -479,7 +479,7 @@ const TreeImageScrubber = (props) => {
               className={classes.navbar}
               buttons={[
                 <Button
-                  variant="outlined"
+                  variant="text"
                   color="primary"
                   onClick={handleFilterClick}
                   startIcon={<IconFilter />}
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -10,9 +10,13 @@ import { makeStyles } from '@material-ui/core/styles';
 import IconLogo from './IconLogo';
 import Menu from './common/Menu.js';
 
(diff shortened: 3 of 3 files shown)


src/components/Planters.js
```diff
@@ -230,6 +230,7 @@ const Planters = (props) => {
             buttons={[
               <IconButton onClick={handleFilterClick} key={1}>
```
nmcharlton, line 231: This would be better as a regular `Button` with a `startIcon`.
See https://material-ui.com/components/buttons/#buttons-with-icons-and-label.

src/components/FilterTop.js
```diff
@@ -144,196 +144,199 @@ function Filter(props) {
...
-                margin="normal"
-                id="start-date-picker"
-                label="Start Date"
-                format={getDateFormatLocale(true)}
-                value={dateStart}
-                onChange={handleDateStartChange}
-                maxDate={dateEnd}
-                KeyboardButtonProps={{
-                  'aria-label': 'change date',
+        <form onSubmit={handleSubmit}>
```
nmcharlton, line 147: The `form` approach is causing the page to reload for me.

https://user-images.githubusercontent.com/5558838/117713347-78523e80-b1cd-11eb-8e85-edc1c2a511a9.mov

src/components/TreeImageScrubber.js
```diff
@@ -480,6 +480,7 @@ const TreeImageScrubber = (props) => {
               buttons={[
                 <IconButton onClick={handleFilterClick} key={1}>
```
nmcharlton, line 481: This would be better as a regular `Button` with a `startIcon`.
See https://material-ui.com/components/buttons/#buttons-with-icons-and-label.

nmcharlton (changes requested):

src/components/TreeImageScrubber.js
tranquanghuy0801: I have fixed those issues in my latest commit.

nmcharlton: Thanks @tranquanghuy0801.
I hope you don't mind, I've made a couple of tweaks to the styling to centre the button vertically and remove the outline.

nmcharlton (approved):

nmcharlton:
Resolves #29 
Resolves #80

PR #87, opened 2021-05-16
fix: reduce # of requests to server and add conditional rendering

gwynndp:
TreeImageScrubber.js
	- manage rendering of PlanterDetail, EditPlanter, and FilterTop until user opens them 
	- consolidate useEffects to reduce number of requests to server
	- refactor so that organizations are loaded once instead of by each sub-component that uses them

FilterTop.js & EditPlanter.js
	- refactor to use organizationList from organization store
	- conditionally render the Organization dropdown based on if the user is signed into an org account

PlanterDetail.js
	- conditionally render EditPlanter based on if the user has opened it or not

### DIFF
--- a/src/components/Verify.js
+++ b/src/components/Verify.js
@@ -866,7 +866,7 @@ function SidePanel(props) {
                   onClick={() => setRejectionReason('unapproved_tree')}
                   value="unapproved_tree"
                   control={<Radio />}
-                  label="Not an approved capture"
+                  label="Not an approved tree"
                 />
                 <FormControlLabel
                   onClick={() => setRejectionReason('blurry_image')}


src/components/FilterTop.js
```diff
@@ -84,14 +87,23 @@ function Filter(props) {
...
   }, [tagSearchString, props.tagsDispatch]);
 
   useEffect(() => {
-    props.organizationDispatch.loadOrganizations();
-  }, [props.organizationDispatch]);
+    log.debug('filter top checks user org id & loads orgs');
+    const hasOrg = getOrganization();
+    setUserHasOrg(hasOrg ? true : false); // check if it's an org account or admin
+    if (!hasOrg) {
+      setOrgList(props.organizationState.organizationList); // only load if it's not an org account
```
nmcharlton, line 104: We're still erroneously loading the full org list into the client, just not into the `FilterTop` component.

nmcharlton (changes requested):
Similar to my comments on Greenstand/treetracker-admin-api#562, we need to disable the org list request for org users until we have restricted org queries implemented.

src/components/TreeImageScrubber.js
```diff
@@ -609,11 +602,13 @@ const TreeImageScrubber = (props) => {
             className={classes.snackbar}
           />
         )}
-      <PlanterDetail
-        open={planterDetail.isOpen}
-        planterId={planterDetail.planterId}
-        onClose={() => handleClosePlanterDetail()}
-      />
+      {planterDetail.isOpen && (
```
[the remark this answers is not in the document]
gwynndp: When it's gotta go, it's gotta go Nick!  --- Ok, I'll remove conditional rendering :)

src/components/FilterTop.js
gwynndp: Working to resolve on the server...

src/api/treeTrackerApi.js
```diff
@@ -368,18 +368,23 @@ export default {
       .catch(handleError);
   },
   /*
-   * get organizations
+   * get organizations for
    */
   getOrganizations() {
-    const query = `${process.env.REACT_APP_API_ROOT}/api/organizations?filter[where][type]=o`;
+    // get all the organizations that the user belongs to? or get all organizations?
+    const query = `${process.env.REACT_APP_API_ROOT}/api/organizations?filter[where][type]=O`;
```
gwynndp, line 375: @nmcharlton 
Edited to search for capital "O" organization type

src/components/FilterTop.js
gwynndp: I left this state setup because I believe it may be helpful when we want to include/exclude sub organizations in future.

src/components/TreeImageScrubber.js
```diff
@@ -206,33 +207,25 @@ const TreeImageScrubber = (props) => {
...
+    document.title = `Verify - ${documentTitle}`;
     props.verifyDispatch.updateFilter(props.verifyState.filter);
     props.verifyDispatch.loadTreeImages();
-  }, [props.verifyDispatch]);
-
-  /* to update html document title */
-  useEffect(() => {
-    document.title = `Verify - ${documentTitle}`;
+    props.verifyDispatch.getTreeCount();
+    props.organizationDispatch.loadOrganizations();
```
nmcharlton, line 214: Triggering this loading here rather than in the components that use it means that `organizationList` is not necessarily populated when the component loads. In particular, you would need to open the _Verify_ page before you navigate to _Edit Planter_.

nmcharlton (changes requested):
Just one more comment about how we load the organization list

src/components/TreeImageScrubber.js
gwynndp: True.  I see the reasoning for having each component load the state they need, especially if they are loaded from different places.
For the verify page there are many requests all happening when that page loads,  were two components making the same request and they are both updating a central state in the verify store.  I believe there are only two paths to load Edit Planter, through the verify page and through the planters page -- is there another route?

Perhaps, especially with the organization list, it would make sense to load this data when the user logs in, then the sub-components would make direct requests to load only if that state changed?

gwynndp: True. I see the reasoning for having each component load the state they need, especially if they are loaded from different places.  

For the verify page, there are many requests all happening when that page loads, all updating the central stores.  Filter Top and the Edit Planter were duplicating the request for organization list and looked like they might be causing each other to re-render when that value updated in the store. (There are still random re-renders, possibly related to the species requests instead of the org list)

I can step back to components making the requests and see if I can manage the triggers to render based on checking the store for the values first.

With the organization list, it might make sense to load this data when the user logs in, set the store, then any sub-components relying on the data would make direct requests to load only if that state changed?

What do you think?

nmcharlton: I agree,loading once on login would make sense.

gwynndp: Moved loading organizations to Account page which is first to load after login.  Is there a better place to put it?

nmcharlton: I don't think it makes sense to put the request in another component that doesn't use organizations.
I would suggest leaving it in the components that use it, but if you want to reduce duplication of API requests, having the model check whether it needs to make the request.

src/components/Verify.js
```diff
@@ -857,7 +866,7 @@ function SidePanel(props) {
                   onClick={() => setRejectionReason('unapproved_tree')}
                   value="unapproved_tree"
                   control={<Radio />}
-                  label="Not an approved tree"
+                  label="Not an approved capture"
```
nmcharlton, line 869: My interpretation of this rejection reason is that the type of tree being captured is not approved for inclusion, for example an invasive species.

nmcharlton (approved):
Great work @gwynndp, thanks for all your efforts on this

PR #89, opened 2021-05-17
fix: Approve UX 

### LINKED ISSUE #11: UX issue: It's not clear which tags are saved on tree rejection
The *Tags* sidebar on the *Verify* page has a set of tag options, some of which fall under *Approve* and *Reject* headings.

What's not clear is that when you reject a tree, only the `rejection_reason` is recorded.
All other tags (`morphology`, `age` and `species_id`) are discarded, even though the inputs are still enabled and visible.

Perhaps all tag inputs should sit under either *Approve* or *Reject*.

tags

reject

luacmartins:
Resolves #11.

@nmcharlton Is this the type of functionality we are looking for? If not, I will be glad to make changes.

### DIFF
--- a/..env.development.local
+++ b/..env.development.local
@@ -1,2 +1,0 @@
-REACT_APP_WEBMAP_DOMAIN=http://dev.treetracker.org
-REACT_APP_API_ROOT=http://localhost:3000
--- a/src/components/TreeImageScrubber.js
+++ b/src/components/TreeImageScrubber.js
@@ -170,6 +170,9 @@ const useStyles = makeStyles((theme) => ({
   sidePanelSubmitButton: {
     width: '128px',
   },
+  mb: {
+    marginBottom: '1rem',
+  },
 }));
 
 const ToVerifyCounter = withData(({ data }) => (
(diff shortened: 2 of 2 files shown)


..env.development.local
```diff
@@ -0,0 +1,2 @@
+REACT_APP_WEBMAP_DOMAIN=http://dev.treetracker.org
```
nmcharlton, line 1: I don't think you meant to commit this 😉

src/components/TreeImageScrubber.js
```diff
@@ -767,59 +710,135 @@ function SidePanel(props) {
...
+          </RadioGroup>
+        </Grid>
+        */}
+              <Grid>
+                <Typography className={classes.sidePanelItem} variant="h6">
+                  Species (if known)
+                </Typography>
+                <Species />
+              </Grid>
+              <Grid>
```
nmcharlton, line 776: The _Additional tags_ should be available for both _Approve_ and _Reject_ operations.

nmcharlton:
> @nmcharlton Is this the type of functionality we are looking for? If not, I will be glad to make changes.

@luacmartins Yep, this is exactly what I had in mind 😃

..env.development.local
luacmartins: Oops! Good catch, thanks!

src/components/TreeImageScrubber.js
luacmartins: Makes sense. I added them to both now!

nmcharlton (approved):
Perfect, thanks @luacmartins!

Issue #86, opened 2021-05-10
Dash Stats: Total Trees count is incorrect for organization users


nmcharlton:
Log in to the Admin Panel as an organization user (e.g. one with the _Freetown Manager_ role).
_Untagged Trees_ plus _Verified Trees_ should equal _Total Trees_.

Instead, _Total Trees_ is reporting the same count for normal and organization users:
![Screenshot 2021-05-10 at 19 43 05](https://user-images.githubusercontent.com/5558838/117708618-f14e9780-b1c7-11eb-9117-e35156362917.png)

tranquanghuy0801:
HI @nmcharlton , can I have a look at this issue?

nmcharlton:
This issue is related to #28.

Issue #28, opened 2021-02-24
Dash stats incorrectly reflect filters selected in Verify and Planters tools


nmcharlton:
The _Untagged Trees_ and _Planters_ dashboard cards should display the totals available to the user (i.e. restricted to organization trees and planters for organization users).

![Screenshot 2021-02-24 at 22 17 08](https://user-images.githubusercontent.com/5558838/109073258-0ac66400-76ee-11eb-8c6c-c625571c59c3.png)

There is a bug whereby those numbers change to reflect filters applied in the Verify and Planters tools respectively.
This is because they share a model. The dashboard should probably have its own model to avoid interference by other components.

![Screenshot 2021-02-24 at 22 21 46](https://user-images.githubusercontent.com/5558838/109073715-afe13c80-76ee-11eb-90c7-34504d976edd.png)

PR #92, opened 2021-05-23
Fix dashboard displaying total tree counts

tranquanghuy0801:
Resolves issue #86

### DIFF
--- a/src/components/DashStat.container.js
+++ b/src/components/DashStat.container.js
@@ -27,8 +27,10 @@ export const DashStatTotalTrees = compose(
   connect(
     (state) => ({
       data:
-        state.trees.treeCount !== null
-          ? countToLocaleString(state.trees.treeCount)
+        state.verify.treeCount + state.verify.verifiedTreeCount !== null
+          ? countToLocaleString(
+              state.verify.treeCount + state.verify.verifiedTreeCount,
+            )
           : null,
       needsRefresh: true,
     }),


src/components/DashStat.container.js
```diff
@@ -27,8 +27,10 @@ export const DashStatTotalTrees = compose(
   connect(
     (state) => ({
       data:
-        state.trees.treeCount !== null
-          ? countToLocaleString(state.trees.treeCount)
+        state.verify.treeCount + state.verify.verifiedTreeCount !== null
```
nmcharlton, line 30: This null check doesn't work – the `+` operator always returns a valid number by interpreting `null` as `0`.
You can probably just remove this check.

src/components/DashStat.container.js
```diff
@@ -27,8 +27,10 @@ export const DashStatTotalTrees = compose(
...
+          ? countToLocaleString(
+              state.verify.treeCount + state.verify.verifiedTreeCount,
```
nmcharlton, line 32: `state.verify.treeCount` will change when filters are set in the _Verify_ page, making this total no longer valid (see #28).
Also, this total will not include rejected captures by default, meaning this total will only include unverified and approved captures.

Instead, we need a steady value for each of these stats, like `state.verify.unverifiedTreeCount`.

nmcharlton (changes requested):
It may be worth fixing #28 at the same time as this one, since they're closely related.

nmcharlton:
This PR has been superseded by #107

Issue #53, opened 2021-03-26
Add Tokenized to Verify/Trees filter and Tree Detail


nmcharlton:
If a capture has a non-null `token_id` in the database, it is considered _Tokenized_.

We need to display whether or not a capture is Tokenized (in Tree Detail) and allow the user to filter on this value in the Verify and Trees pages (see #51 which aims to consolidate the filters on the two pages).

nmcharlton:
This will require a supporting change in [treetracker-admin-api](https://github.com/Greenstand/treetracker-admin-api).

VLuisa:
I can pick this one up!

VLuisa:
@nmcharlton I'm less familiar with how Redux is set up in the app and how to go about adding in a new ajax request to get access to the token_id data (not sure if I'm even approaching this correctly) - mind giving me some pointers on how to get the right query for the token_id built in and where to do that? I'm wondering whether it makes sense to create two queries, one that looks for all null tokenIds and one that looks for all non-null

VLuisa:
@nmcharlton Ah ok, makes sense. I was definitely overcomplicating it! Thanks!

Issue #51, opened 2021-03-26
Make Verify and Captures pages more consistent


nmcharlton:
The Captures page has fallen out of sync with the Verify page.
The following steps should be taken to make these more consistent.

## Capture detail
Use the same Capture Detail popup for the two pages (the Verify one is best)

## Filters
Provide the same set of filters and filtering experience.
Ideally, this should be the same component with responsive sizing to adapt to top vs side layouts.
Again, the Verify filter is more appropriate (with a Reset/Clear button as proposed by @gwynndp in #38)

## Captures page columns
Change the columns on Captures to match the details listed in Capture Detail.

nmcharlton:
See #10 for better Capture Detail dialog design

tranquanghuy0801:
Hi @nmcharlton , can I work on this issue?

nmcharlton:
@tranquanghuy0801 Let me know if you think it's worth splitting this issue out into multiple issues to make it easier to implement part of it.

tranquanghuy0801:
@nmcharlton 
I think there is no need to split this issue at the moment. If I struggle on any parts of this, I will let you know.

tranquanghuy0801:
@nmcharlton, is that compulsory for making the Captures table have the same data as the Capture Detail dialog as there are a lot of data to fill in the Captures table?

PR #101, opened 2021-05-30
Better capture details & filters

### LINKED ISSUE #54: Refine and restructure fields in Verify/Trees filter, Trees list and Tree Detail
The fields currently displayed for a capture are basically mirrors of the database.
Instead they should display information in a format that is meaningful and logical for the user.

For example, we use two Boolean fields (`active` and `approved`) to denote the verification state of a capture:

- `active===true && approved===false` => Awaiting verification
- `active===true && approved===true` => Approved
- `active===false && approved===false` => Rejected

These should be combined into one _Verification State_ field for display and filter.

Data in the Tree Detail could be grouped by:

- Capture data (Planter ID, Planter Identifier, Device Identifier, Date Created, Note)
- Verification state
- Capture tags (Tags, Species)
- Tokenized

This will require some UI/UX design effort and input from @sebastiangaertner and other users.

Note that there are several other issues related to the Filters and Tree Detail: #10, #26, #29, #35, #37, #50, #51, #53

Current Tree Detail UI:
Screenshot 2021-03-26 at 22 40 36

VLuisa:
Resolves #54 

* Restructured fields in Capture Details view into groups
* Created labels to consolidate verification status into 'Approved', 'Awaiting Verification', or 'Rejected'.
image
image
image

* Added logical 'Verification Status' filter to replace separate 'Approved' and 'Active' filters:
image
image

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -159,7 +159,6 @@ function CaptureDetailDialog(props) {
               value: capture.deviceIdentifier,
               copy: true,
             },
-            { label: 'Species', value: species && species.name },
             { label: 'Created', value: dateCreated.toLocaleString() },
             { label: 'Note', value: renderCapture.note },
           ].map((item) => (
@@ -205,6 +204,18 @@ function CaptureDetailDialog(props) {
         <Divider />
         <Grid item>
           <Typography className={classes.subtitle}>Tags</Typography>
+          <Typography variant="subtitle1">Species</Typography>
+          {species && species.name ? (
+            <Chip
+              key={species && species.name}
+              label={species && species.name}
+              className={classes.chip}
+            />
+          ) : (
+            <Typography variant="body1">---</Typography>
+          )}
+
+          <Typography variant="subtitle1">Other</Typography>
           {allTags.length === 0 ? (
             <Typography variant="body1">---</Typography>
           ) : (
@@ -219,9 +230,7 @@ function CaptureDetailDialog(props) {
         <Grid item>
           <Typography className={classes.subtitle}>Capture Token</Typography>
           <Typography variant="body1">
-            {capture.token != null
-              ? 'Impact token issued'
-              : 'Impact token not issued'}
+            {getTokenStatus(capture.tokenId)}
           </Typography>
         </Grid>
         <Snackbar
@@ -296,5 +305,15 @@ const mapState = (state) => ({
 const mapDispatch = (dispatch) => ({
   captureDetailDispatch: dispatch.captureDetail,
 });
+
+const getTokenStatus = (tokenId) => {
+  if (tokenId === undefined) {
+    return 'Impact token status unknown';
+  } else if (tokenId === null) {
+    return 'Impact token not issued';
+  } else {
+    return 'Impact token issued';
+  }
+};
 
 export default compose(connect(mapState, mapDispatch))(CaptureDetailDialog);


VLuisa:
@nmcharlton  
Could you clarify the expected changes to the 'Trees list' in the ticket title of #54?

Also per our conversation I created a section to display the token status, but it seems from #53 that it needs an api change. Let me know if I should leave the code as is (checking for null value) or if I should revert that part of the updates.
Thanks!

src/components/CaptureDetailDialog.js
```diff
@@ -116,49 +139,65 @@ function CaptureDetailDialog(props) {
...
+              label: 'Planter Identifier',
+              value: capture.planterIdentifier,
+              copy: true,
+            },
+            {
+              label: 'Device Identifier',
+              value: capture.deviceIdentifier,
+              copy: true,
+            },
+            { label: 'Species', value: species && species.name },
```
nmcharlton, line 161: _Species_ should be part of the _Tags_ group, if possible

src/components/CaptureDetailDialog.js
```diff
@@ -116,49 +139,65 @@ function CaptureDetailDialog(props) {
...
+                  )}
+                </Typography>
+              </Grid>
+            </Fragment>
+          ))}
+        </Grid>
+        <Divider />
+        <Grid item>
+          <Typography className={classes.subtitle}>
+            Verification State
```
nmcharlton, line 185: Let's make this _Verification Status_ for consistency with the filters.

src/components/CaptureDetailDialog.js
```diff
@@ -116,49 +139,65 @@ function CaptureDetailDialog(props) {
...
+          </Typography>
+          {!capture.approved && capture.active ? (
```
nmcharlton, line 187: I'd suggest creating a small helper function to translate `(active, approved)` into a verification status to prevent duplication of this logic and make the relationship more obvious.

src/components/CaptureDetailDialog.js
```diff
@@ -169,6 +208,15 @@ function CaptureDetailDialog(props) {
             </div>
           )}
         </Grid>
+        <Divider />
+        <Grid item>
+          <Typography className={classes.subtitle}>Capture Token</Typography>
+          <Typography variant="body1">
+            {capture.token != null
```
nmcharlton, line 215: Assuming the standard LoopBack pattern is followed, this will be `capture.tokenId` (from the `token_id` column in the `trees` table).
It might be best to check for `null` and `undefined` separately:
- `capture.tokenId === undefined` => Impact token status unknown
- `capture.tokenId === null` => Impact token not issued

That will prevent misleading information if this is released before the API change is made.

src/components/Filter.js
```diff
@@ -210,55 +210,44 @@ function Filter(props) {
...
-      <TextField
-        select
-        value={
-          active === undefined ? 'All' : active === true ? 'false' : 'true'
-        }
-        InputLabelProps={{
-          shrink: true,
-        }}
-        onChange={(e) =>
+              : e.target.value === 'Awaiting Verification' ||
```
nmcharlton, line 229: It's best to create constant variables containing these labels and compare those.
That reduces the risk of a typo breaking something that won't be caught by the linter or interpreter.

nmcharlton (changes requested):
This looks great, just a couple of comments so far to reduce code/logic duplication.

> Could you clarify the expected changes to the 'Trees list' in the ticket title of #54?

The columns in the newly renamed _Captures_ table should match the fields in the capture detail dialog.
We should also us the same `Filter` component for both _Verify_ and _Captures_ (which in turns leads me to think what is now the _Verify_ tool should just be an alternative view in the _Captures_ page and _Verify_ should be redesigned to be a simpler tool exclusively for verification, but that's well outside the scope of this issue).

> Also per our conversation I created a section to display the token status, but it seems from #53 that it needs an api change. Let me know if I should leave the code as is (checking for null value) or if I should revert that part of the updates.

I've suggested distinguishing between `null` and `undefined` to cope with this.
The API change is fairly trivial – I'll get it done this week and will let you know when it's merged to help your testing.

VLuisa:
Thanks for all the great feedback/tips @nmcharlton ! I noticed that using the same component for filtering and updating the columns were part of #51, what if I leave the changes to the column names for that issue?

nmcharlton (approved):
Fantastic work, thanks @VLuisa!

PR #105, opened 2021-06-03
fix: add org id to planter request

gwynndp:
planters.js
- add organization id to getPlanterRegistration request

FilterTop.js
- since the results are being preserved on the page, preserve the filter selections in the filter view too
- TODO: preserve the state of the tag (the AutoComplete component takes an object as a value while the filter preserves only the tag id #)

Verify.js
- remove organizations completely
- remove unnecessary useEffect (getCaptureCount doesn't need to be called again because the count won't be updated by any action that can be taken on this page)

### DIFF
--- a/src/api/planters.js
+++ b/src/api/planters.js
@@ -66,7 +66,9 @@ export default {
   },
 
   getPlanterRegistrations(planterId) {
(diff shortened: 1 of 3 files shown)


nmcharlton (approved):
Looks good 👍

PR #107, opened 2021-06-06
fix: resolve various dash stat problems

### LINKED ISSUE #28: Dash stats incorrectly reflect filters selected in Verify and Planters tools

### LINKED ISSUE #61: Dashboard stats missing on first load after login
There appears to be a race condition with the first API calls following user login, whereby some calls have an undefined Authorization token in the header.

The result is some missing numbers on the dashboard.

Refreshing the page or navigating to another page in the Admin Panel forces a reload of the data, and subsequent sessions while the user is already logged in are fine.

### LINKED ISSUE #86: Dash Stats: Total Trees count is incorrect for organization users

nmcharlton:
Resolves #28 
Resolves #61 
Resolves #86 

Management of all capture-related dash stats has moved to the `captures` model with dedicated state variables for the counts, not affected by filters.
Similarly, the `planters` model has a dedicated total count used in the dash stats.

The session token is now being set before components are notified of a change, meaning there are no errors on first load after login.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "treetracker-admin-client",
-  "version": "1.7.1",
+  "version": "1.13.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/common/numbers.js
+++ b/src/common/numbers.js
@@ -1,4 +1,4 @@
 export function countToLocaleString(num) {
-  if (!num.toLocaleString || typeof num !== 'number') return num;
+  if (!num?.toLocaleString || typeof num !== 'number') return num;
   return num.toLocaleString();
 }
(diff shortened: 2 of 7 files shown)


gwynndp (approved):
Looking good!  Nice pruning

PR #112, opened 2021-06-10
Feature: Tokenized filter option

### LINKED ISSUE #53: Add Tokenized to Verify/Trees filter and Tree Detail

VLuisa:
Resolves #53 

Add 'Tokenization Status' filter to Verify and Capture pages with options to filter by 'All', 'Tokenized', and 'Not Tokenized'.

### DIFF
--- a/src/components/Filter.js
+++ b/src/components/Filter.js
@@ -254,24 +254,12 @@ function Filter(props) {
           </MenuItem>
         ))}
       </TextField>
-      <GSInputLabel text="Tokenization Status" />
+      <GSInputLabel text="Token Status" />
       <TextField
         select
-        value={
-          tokenId === filterOptionAll
-            ? filterOptionAll
-            : tokenId === tokenizationStates.TOKENIZED
-            ? tokenizationStates.TOKENIZED
-            : tokenizationStates.NOT_TOKENIZED
-        }
+        value={tokenId}
         onChange={(e) => {
-          setTokenId(
-            e.target.value === filterOptionAll
-              ? filterOptionAll
-              : e.target.value === tokenizationStates.TOKENIZED
-              ? tokenizationStates.TOKENIZED
-              : tokenizationStates.NOT_TOKENIZED,
-          );
+          setTokenId(e.target.value);
         }}
       >
         {[
--- a/src/components/FilterTop.js
+++ b/src/components/FilterTop.js
@@ -204,21 +204,9 @@ function Filter(props) {
               <TextField
                 select
                 label="Token Status"
-                value={
-                  tokenId === filterOptionAll
-                    ? filterOptionAll
-                    : tokenId === tokenizationStates.TOKENIZED
-                    ? tokenizationStates.TOKENIZED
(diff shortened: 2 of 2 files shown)


src/common/variables.js
```diff
@@ -9,3 +9,7 @@ export const verificationStates = {
   AWAITING: 'Awaiting Verification',
   REJECTED: 'Rejected',
 };
+export const tokenizationStates = {
```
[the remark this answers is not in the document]
VLuisa: Good point! Just addressed it

src/components/Filter.js
```diff
@@ -251,6 +254,36 @@ function Filter(props) {
           </MenuItem>
         ))}
       </TextField>
+      <GSInputLabel text="Tokenization Status" />
+      <TextField
+        select
+        value={
```
nmcharlton, line 260: This can be simplified to `value={tokenId}`.

src/components/Filter.js
```diff
@@ -251,6 +254,36 @@ function Filter(props) {
...
+          tokenId === filterOptionAll
+            ? filterOptionAll
+            : tokenId === tokenizationStates.TOKENIZED
+            ? tokenizationStates.TOKENIZED
+            : tokenizationStates.NOT_TOKENIZED
+        }
+        onChange={(e) => {
+          setTokenId(
```
nmcharlton, line 268: This can be simplified to `setTokenId(e.target.value)`.

src/components/FilterTop.js
```diff
@@ -198,6 +201,36 @@ function Filter(props) {
                   </MenuItem>
                 ))}
               </TextField>
+              <TextField
+                select
+                label="Token Status"
+                value={
```
nmcharlton, line 207: This can be simplified to `value={tokenId}`.

src/components/FilterTop.js
```diff
@@ -198,6 +201,36 @@ function Filter(props) {
...
+                  tokenId === filterOptionAll
+                    ? filterOptionAll
+                    : tokenId === tokenizationStates.TOKENIZED
+                    ? tokenizationStates.TOKENIZED
+                    : tokenizationStates.NOT_TOKENIZED
+                }
+                onChange={(e) => {
+                  setTokenId(
```
nmcharlton, line 215: This can be simplified to `setTokenId(e.target.value)`.

nmcharlton (commented):
I've suggested a couple of code simplifications that take advantage of the recent change.

src/components/FilterTop.js
VLuisa: Heh whoops, thanks for catching this!

nmcharlton (approved):
Cracking work, thanks @VLuisa!

PR #113, opened 2021-06-10
fix: issue_94

### LINKED ISSUE #94: Filter date selection: Cannot move forward
The forward button on the date selector used in filters on the _Verify_ and _Trees_ pages is permanently disabled, meaning you can't navigate to the next month if you go back.
Instead, you need to close the date selector and reopen it.

https://user-images.githubusercontent.com/5558838/119395353-4a8bf000-bccb-11eb-9ed6-e6043b200caf.mov

panzaverdeok:
Resolves #94

### DIFF
--- a/src/components/FilterTop.js
+++ b/src/components/FilterTop.js
@@ -214,7 +214,7 @@ function Filter(props) {
                   format={getDateFormatLocale(true)}
                   value={dateStart}
                   onChange={handleDateStartChange}
-                  maxDate={dateEnd}
+                  //maxDate={dateEnd} Conrado Banegas comment 6/2021 issue #94
                   KeyboardButtonProps={{
                     'aria-label': 'change date',
                   }}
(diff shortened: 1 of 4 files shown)


src/components/FilterTop.js
```diff
@@ -214,7 +214,7 @@ function Filter(props) {
                   format={getDateFormatLocale(true)}
                   value={dateStart}
                   onChange={handleDateStartChange}
-                  maxDate={dateEnd}
+                  //maxDate={dateEnd} Conrado Banegas comment 6/2021 issue #94
```
nmcharlton, line 217: A few comments:

1. We shouldn't remove `maxDate`, but should set it to today's date if `dateEnd` is `null` or `undefined`.
2. The same problem exists with `minDate` on the End Date `KeyboardDatePicker` (line 229). When `dateStart` is `null` or `undefined`, this should be set to a sensible minimum (e.g., `Date(1900-01-01)`).
3. If code is not needed, don't comment it out – remove it. We can use git to see what has been removed and put it back in if necessary.
4. There's no need to add comments like this to the code. After a while they get in the way of the code and cause the files to bloat. GitHub tracks line-by-line changes for us and we link commits to issues via PRs.

nmcharlton (changes requested):
@panzaverdeok I've requested some changes. Please ask if you want some help with them.

src/components/FilterTop.js
panzaverdeok: Hi Nick.
It' ok, I'll try to develop another alternativa to solve the issue.
Thanks.

PR #118, opened 2021-06-17
refactor: move redux to context

gwynndp:
Refactored organization and planter states from redux to Context.js

Incomplete: Dashboard containers refactored to remove redux but no replacement data available so hard-coded to "111" execpt for the planters #

NOTE: "--" before a filename means I think I've removed it's use from the code but I'm temporarily keeping it for reference.

### DIFF
--- a/src/components/FilterTopPlanter.js
+++ b/src/components/FilterTopPlanter.js
@@ -89,7 +89,7 @@ function FilterTopPlanter(props) {
   };
 
   const handleEnterPress = (e) => {
-    e.key === 'Enter' && handleSubmit();
+    e.key === 'Enter' && handleSubmit(e);
   };
 
   return (
@@ -193,59 +193,10 @@ function FilterTopPlanter(props) {
                 label="Phone Number"
                 htmlFor="Phone Number"
                 id="Phone Number"
-                personId="Phone Number"
                 placeholder="Phone Number"
                 value={phone}
                 onChange={(e) => setPhone(e.target.value)}
                 onKeyDown={handleEnterPress}
-              />
-              <Button
-                className={classes.apply}
-                label="submit"
-                variant="outlined"
-                color="primary"
-                onClick={handleSubmit}
-              >
-                Apply
-              </Button>
-              <TextField
-                className={`${classes.textField} ${classes.filterElement}`}
-                label="First Name"
-                htmlFor="First Name"
-                id="First Name"
-                placeholder="First Name"
-                value={firstName}
-                onChange={(e) => setFirstName(e.target.value)}
-                onKeyDown={handleEnterPress}
-              />
-              <TextField
-                className={`${classes.textField} ${classes.filterElement}`}
-                label="Last Name"
-                htmlFor="Last Name"
-                id="Last Name"
-                placeholder="Last Name"
-                value={lastName}
-                onChange={(e) => setLastName(e.target.value)}
-                onKeyDown={handleEnterPress}
-              />
-              <TextField
-                className={`${classes.textField} ${classes.filterElement}`}
-                label="Email"
-                htmlFor="Email"
-                id="Email"
-                placeholder="Email"
-                value={email}
-                onChange={(e) => setEmail(e.target.value)}
-              />
-              <TextField
-                className={`${classes.textField} ${classes.filterElement}`}
-                label="Phone Number"
-                htmlFor="Phone Number"
-                id="Phone Number"
-                personId="Phone Number"
-                placeholder="Phone Number"
-                value={phone}
-                onChange={(e) => setPhone(e.target.value)}
               />
             </Grid>
             <Grid className={classes.inputContainer}>
@@ -257,7 +208,7 @@ function FilterTopPlanter(props) {
                 id="submit"
                 variant="outlined"
                 color="primary"
-                onClick={(e) => handleSubmit(e)}
+                onClick={handleSubmit}
               >
                 Apply
               </Button>
--- a/src/components/Planters.js
+++ b/src/components/Planters.js
@@ -215,7 +215,7 @@ const Planters = (props) => {
   );
 
   return (
-    <React.Fragment>
+    <>
       <Grid container direction="column" className={classes.outer}>
         <Grid item>
           <Navbar
@@ -275,7 +275,7 @@ const Planters = (props) => {
         planterId={planterDetail.id}
         onClose={() => setDetailShown(false)}
       />
-    </React.Fragment>
+    </>
   );
 };
 
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -1,10 +1,4 @@
-import React, {
-  Fragment,
-  useState,
-  useEffect,
-  useCallback,
-  useContext,
-} from 'react';
+import React, { useState, useEffect, useCallback, useContext } from 'react';
 
 import { makeStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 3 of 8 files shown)


src/components/Planters.js
```diff
@@ -126,62 +128,50 @@ const useStyles = makeStyles((theme) => ({
...
-  const [planterDetail, setPlanterDetail] = React.useState({});
+  const { planters, setPlanters } = useContext(AppContext);
+  const [isFilterShown, setFilterShown] = useState(false);
+  const [isDetailShown, setDetailShown] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filter, setFilter] = useState(new FilterPlanter());
+  const [planterCount, setPlanterCount] = useState(0);
+  const [planterDetail, setPlanterDetail] = useState({});
+  const [pageSize, setPageSize] = useState(24);
+  const [pageNumber, setPageNumber] = useState(0);
```
dadiorchen, line 139: How about a object state:
{
  planterCount: 
  pageSize
  pageNumber
}

Because I suppose these props would be updated together, usually?

src/context/Context.js
```diff
@@ -1,14 +1,16 @@
-import React from 'react';
```
nmcharlton, line 1: Let's rename this to `AppContext.js` for consistency.

src/components/Planters.js
gwynndp: Thanks for that idea!  I'll take a look and see if I can consolidate them.

src/components/Species.js
```diff
@@ -16,37 +16,34 @@ const styles = () => {
...
-    props.speciesDispatch.loadSpeciesList();
-  }, [props.speciesDispatch]);
+    // don't call unless the list is empty
+    if (context.speciesList.length <= 0) {
+      context.loadSpeciesList();
+    }
+  }, []);
 
   return (
     <Autocomplete
```
nmcharlton, line 29: Need to wrap this with `SpeciesProvider`

src/components/CaptureTags.js
```diff
@@ -134,13 +135,13 @@ const CaptureTags = (props) => {
...
 
   let handleDeleteChip = (_chip, index) => {
-    const temp = props.tagState.tagInput;
+    const temp = context.tagInput;
     temp.splice(index, 1);
-    props.tagDispatch.setTagInput(temp);
+    context.setTagInput(temp);
   };
 
   return (
```
nmcharlton, line 147: Need to wrap component with `TagsProvider`

nmcharlton (commented):
On the whole, the refactor looks solid.
Just need to make sure we wrap all components that use contexts with the corresponding providers.

As for the model tests, we should be able to port them across to the corresponding contexts.

src/components/Navbar.js
```diff
@@ -48,9 +48,7 @@ const Navbar = (props) => {
           <Grid item>{props.children}</Grid>
         </Grid>
       </AppBar>
-      {/* children duplicated behind the AppBar component to preserve height */}
       <Toolbar className={classes.toolbar} />
-      {props.children}
```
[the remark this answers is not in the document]
gwynndp: Good points!  I'll put back the hack for now.  I think now that we're using context it isn't triggering multiple requests, but I'll look for another solution.

src/context/CapturesContext.js
```diff
@@ -0,0 +1,362 @@
...
+  };
+
+  getCapturesAsync = async (filterInfo) => {
+    // Destruct payload and fill in any gaps from state.captures
+    const { page, rowsPerPage, filter, orderBy, order } = {
+      ...this.state,
+      ...filterInfo,
+    };
+    // first load the page count
+    if (!this.state.captureCount) {
```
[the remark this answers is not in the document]
gwynndp: Thanks, I removed it now :)

src/components/DashStat.container.js
```diff
@@ -10,76 +10,87 @@ import LocalOfferOutlinedIcon from '@material-ui/icons/LocalOfferOutlined';
...
-import { AppContext } from '../context/Context';
+import apiPlanters from '../api/planters';
+import api from '../api/treeTrackerApi';
+import FilterModel from '../models/Filter';
+import { AppContext } from '../context/AppContext';
 
 function DashStatTotalCaptures(props) {
-  // data ===> state.captures.totalCaptureCount
-  // const { totalPlanterCount, setTotalPlanterCount } = useContext(AppContext);
+  const defaultFilter = new FilterModel({});
```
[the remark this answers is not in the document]
gwynndp: Do we also care about the value of "approved" for this filter?

src/components/DashStat.container.js
```diff
@@ -10,76 +10,87 @@ import LocalOfferOutlinedIcon from '@material-ui/icons/LocalOfferOutlined';
...
 
-  // useEffect(() => {
-  //   getTotalPlanterCount();
-  // }, []);
+  const [total, setTotal] = useState(0);
```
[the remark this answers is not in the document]
gwynndp: Ok, I updated all the totals to null as initial state.  We don't have a loading state here yet, so that would need a new issue

Mloweedgar (approved):

src/components/FilterTopPlanter.js
```diff
@@ -3,8 +3,10 @@ import { withStyles } from '@material-ui/core/styles';
 import Grid from '@material-ui/core/Grid';
 import Button from '@material-ui/core/Button';
 import TextField from '@material-ui/core/TextField';
+// import MenuItem from '@material-ui/core/MenuItem';
 import FilterModel from '../models/FilterPlanter';
-import { connect } from 'react-redux';
+// import { ALL_ORGANIZATIONS, ORGANIZATION_NOT_SET } from '../models/Filter';
```
gwynndp, line 8: Commented out code is an incomplete refactor to offer the same Organization dropdown that is on other filters.  There is a bug when it's implemented: not all the suborganizations seem to get get included when Freetown is selected from the dropdown.

src/context/CaptureDetailContext.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const getSpecies = async (speciesId) => {
+    if (speciesId == null) {
+      return Promise.resolve(STATE_EMPTY.species);
+    }
+
+    return api.getSpeciesById(speciesId).then((species) => {
+      setState({ ...state, species });
+    });
+  };
```
dadiorchen, line 63: @gwynndp I am not sure I'm correct, maybe context approach has some downsides like: if using Redux, we can reuse this piece of code by put it into the species module/action/reducer, so we can invoke species.dispatch.getSpeciesByID here?

dadiorchen (commented):
@gwynndp Great job, I don't have too much suggestion now, just couple of question:
* What's your feeling on writing test with testing-library, I'm a little scared about writing pure test code against UI/components, cuz I must to rely on command to check the result. I'm considering using Cypress unit + integration test to test the UI, it's easier because you can check the test in browser and refresh it. Then we use headless command to run it in CI, of cause the downsides might be slower performance.
* So mainly, in the future, the business logic code previously in the Redux folder would be located in the context folder in different context files respectively, right? So the file like verifyContext would be a bit of like the model layer, am I correct?
* What's the goal of the views?

dadiorchen (approved):

src/components/tests/organizations.test.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
```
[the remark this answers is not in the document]
gwynndp: I've found it to be a learning curve but I think most of the difficulty is because our components are too big and integrated to begin with.  It will work better with smaller components that are more independent and modular.

It does seem to be harder to test the business logic in the context, but I think that is just a different learning curve that will involve mocking child components.  Most of the logic in the admin panel boils down to "did the api request work", so I think there may be other ways to test that than react-testing-library.  Regular Jest tests are still compatible.

We are also not planning on having these tests replace the cypress integration tests.  Once the components are broken down these tests should work more as modular/unit tests for development.  As it is I can see some overlap between the testing-library tests and the ones we have in cypress.  

I do like that these tests will be independent of code implementation once they're implemented well.  

For the meantime I aim to try breaking down the tests and organizing them by component and improve the cypress integration tests as well.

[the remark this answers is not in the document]
gwynndp: Yes, the contexts are currently replacing the models for our state management.

The views are to help organize the code into component trees that will share the same state context.  They already existed as wrapper components that the router to reference to render each page in the admin panel.  We just wanted to put them together for consistency, and because it'll be easier to add context and any other logic needed by the child components at those root files, etc.

.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
+{
```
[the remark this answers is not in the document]
gwynndp: This can be removed.  I was trying some settings for testing the tests using VSCode's debugger.

src/context/CaptureDetailContext.js
gwynndp: Context and redux do have different strengths and weaknesses. Context is good for holding state that might be used by a particular tree of components.  The .getSpeciesById() can be called without context if the api file is imported, so I'm not sure there is a limitation in this case.  But, I do think that we might review the organization of data into the contexts.  I kept it close to the way it was organized in the stores wit the refactor, but it might not hurt to allow verify context to hold data for tags and species if it helps simplify things, for example.   

There are some changes already planned for the admin panel that may make these concerns mute, so we should revisit this discussion later.

PR #124, opened 2021-07-03
fix: support complex image urls

### LINKED ISSUE #109: CDN resize not working when planter image comes from DB directly (not AWS)
See video 
https://youtu.be/V0g94Gc_5wE

nmcharlton:
Resolves #109 

The first group in the regex now does lazy matching so it only extracts the domain from the url.
The full path is therefore left for the greedy second group.

### DIFF
--- a/src/components/OptimizedImage.spec.py.js
+++ b/src/components/OptimizedImage.spec.py.js
@@ -1,0 +1,75 @@
+/* eslint-disable */
+import { mount } from 'cypress-react-unit-test';
+import React from 'react';
+import theme from './common/theme';
+import { ThemeProvider } from '@material-ui/core/styles';
+
+import OptimizedImage from './OptimizedImage';
+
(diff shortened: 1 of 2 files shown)


src/components/OptimizedImage.spec.py.js
```diff
@@ -0,0 +1,75 @@
...
+      cy.get(`img`).and(($img) => {
+        expect($img[0].src).to.contain('cdn.statically.io');
+      });
+    });
+  });
+
+  describe('with complex url', () => {
+    beforeEach(() => {
+      const imageUrl =
+        'https://greenstand.org/fileadmin/_processed_/d/4/csm_little_Jony_ca31f27bd1.jpg';
```
luacmartins, line 51: NAB remove additional empty line.

luacmartins (approved):
Code looks good to me. Great work @nmcharlton!

Issue #95, opened 2021-05-24
Tree ID and Planter ID click in Trees list should not open the tree dialog


nmcharlton:
On the _Trees_ page, clicking the _Tree ID_ or _Planter ID_ opens the web map in a new tab.
Clicking anywhere else on the tree row opens the tree detail.

Clicks on _Tree ID_ and _Planter ID_ should not be passed through to the background and should not open the tree detail.

https://user-images.githubusercontent.com/5558838/119400714-78286780-bcd2-11eb-817e-fc251c8f97f6.mov

nmcharlton:
Fixed in #126

PR #126, opened 2021-07-06
Fix: Click PlanterID would not open the Planter Detail

### LINKED ISSUE #96: Clicking Planter ID on planter card should not open Planter Detail
Clicking a planter card in the _Planters_ page opens the _Planter Detail_ dialog.
Clicking the _ID_ opens the web map for that planter, but also erroneously opens the _Planter Detail_.

The _ID_ clicks should not be passed through to the card.

https://user-images.githubusercontent.com/5558838/119401805-f1748a00-bcd3-11eb-97d7-c1b5d8017317.mov

tranquanghuy0801:
Resolved #96

### DIFF
--- a/src/components/Planters.js
+++ b/src/components/Planters.js
@@ -128,7 +128,6 @@ const Planters = (props) => {
   const classes = useStyles(props);
   const [isFilterShown, setFilterShown] = React.useState(false);
   const [isDetailShown, setDetailShown] = React.useState(false);
-  const [isClickLink, setClickLink] = React.useState(false);
   const [planterDetail, setPlanterDetail] = React.useState({});
 
   /*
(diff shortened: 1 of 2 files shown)


nmcharlton (changes requested):
Although this approach has the desired effect, it is unnecessarily complicated (and therefore more prone to breaking when the code changes) and only fixes one instance.

I'd suggest addressing the problem in the `LinkToWebmap` component by stopping click propagation for the `Link`:
```
  return (
    <Link
      onClick={(e) => e.stopPropagation()}
      href={`${process.env.REACT_APP_WEBMAP_DOMAIN}/?${type}id=${value}`}
      underline="always"
      target="_blank"
    >
      {value}
      <OpenInNewIcon fontSize="inherit" className={classes.openInNewIcon} />
    </Link>
  );
```

This would also address issue #95.

nmcharlton (approved):

PR #127, opened 2021-07-07
feat: add phone and email filters to planters

### LINKED ISSUE #111: Add planter identifier to Planters filter
We need the ability to filter planters by planter identifier (`planter_identifier` in the database and `planterIdentifier` in the API response object). This will either be an email address or a phone number (depending on the planter).

Screenshot 2021-06-09 at 20 01 52

VLuisa:
Resolves #111 

image
image

### DIFF
--- a/src/models/FilterPlanter.js
+++ b/src/models/FilterPlanter.js
@@ -40,7 +40,9 @@ export default class Filter {
     }
 
     if (this.phone) {
-      where.phone = this.phone;
+      where.phone = {
+        ilike: this.phone,
+      };
     }
 
     return where;


src/models/FilterPlanter.js
```diff
@@ -33,6 +33,16 @@ export default class Filter {
...
     }
 
+    if (this.email) {
+      where.email = {
+        ilike: this.email,
+      };
+    }
+
+    if (this.phone) {
+      where.phone = this.phone;
```
nmcharlton, line 43: I think this should use `ilike` as well for consistency with the other string filters.

nmcharlton (changes requested):
Just one suggestion.

src/models/FilterPlanter.js
VLuisa: Great point

nmcharlton (approved):
Nice work 👍

Issue #117, opened 2021-06-15
Unable to edit Species


sebastiangaertner:
Trying to add description to an existing species does not work.

nmcharlton:
Editing an existing species gives the following message:

![Screenshot 2021-06-16 at 20 17 37](https://user-images.githubusercontent.com/5558838/122279435-f0ec9f00-cedf-11eb-926d-dc8c9d05fd63.png)

This suggests that either the check for duplicate species name is in the wrong place, or we're calling the _create species_ function instead of _update species_.

tranquanghuy0801:
Hi, can I work on this issue?

PR #128, opened 2021-07-09
fix: edit species

tranquanghuy0801:
Resolved issue #117

### DIFF
--- a/src/components/SpeciesTable.js
+++ b/src/components/SpeciesTable.js
@@ -374,12 +374,6 @@ const EditModal = ({
   data,
 }) => {
   const [error, setError] = React.useState(undefined);
-  let objSpecies = {};
-  data.map((species) => {
-    if (!objSpecies[species.id]) {
-      objSpecies[species.id] = species.name.toLowerCase();
-    }
-  });
 
   const onNameChange = (e) => {
     setError(undefined);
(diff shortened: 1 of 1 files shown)


src/components/SpeciesTable.js
```diff
@@ -386,12 +391,20 @@ const EditModal = ({
...
 
   const handleEditDetailClose = () => {
+    setError(undefined);
     setIsEdit(false);
     setSpeciesEdit(undefined);
   };
 
   const handleSave = async () => {
-    if (nameSpecies.includes(speciesEdit.name.toLowerCase().trim())) {
+    let editName = speciesEdit.name.toLowerCase().trim();
```
nmcharlton, line 400: We should use `const` instead of `let` if the variable doesn't change after assignment.

src/components/SpeciesTable.js
```diff
@@ -386,12 +391,20 @@ const EditModal = ({
...
+    let nameSpecies = isEdit
```
nmcharlton, line 401: Use `const` instead of `let`.

src/components/SpeciesTable.js
```diff
@@ -386,12 +391,20 @@ const EditModal = ({
...
+      ? Object.keys(objSpecies)
+          .filter((id) => parseInt(id) !== speciesEdit.id)
+          .map((id) => objSpecies[id])
+      : Object.values(objSpecies);
+    let existSpecies = nameSpecies.filter((name) => name === editName);
```
nmcharlton, line 406: Use `const` instead of `let`.

nmcharlton: You could just check `nameSpecies.includes(editName)` which returns a boolean.

src/components/SpeciesTable.js
```diff
@@ -374,7 +374,12 @@ const EditModal = ({
   data,
 }) => {
   const [error, setError] = React.useState(undefined);
-  const nameSpecies = data.map((species) => species.name.toLowerCase());
+  let objSpecies = {};
+  data.map((species) => {
```
nmcharlton, line 378: Populating `objSpecies` seems like unnecessary complexity.
You could just filter the species list here and leave the rest of the code as it is – maybe something like:
```
const otherSpeciesList = isEdit
  ?  data.filter((species) => Number(species.id) !== speciesEdit.id)
  :  data;
const nameSpecies = otherSpeciesList.map((species) => species.name.toLowerCase());
```

nmcharlton (commented):
I've made some suggestions to simplify the implementation and make the code a little clearer.

tranquanghuy0801:
@nmcharlton, thanks for your feedback, I have updated my code

nmcharlton (approved):

PR #134, opened 2021-07-14
feat: [#133] add reset button and enter keydown functionality to planter view filter

### LINKED ISSUE #133: Planter filter: Reset button and Apply on enter
The filter in the Verify tool has two features that would be useful in the Planters filter:

1. A _Reset_ button that sets filters fields back to their default values (empty/null)
2. Pressing _enter_ triggers _Apply_

Verify filter:
Screenshot 2021-07-13 at 08 31 18

Planters filter:
Screenshot 2021-07-13 at 08 31 36

We should strive for consistency across these components.

thejoshcooter:
Resolves #133

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "treetracker-admin-client",
-  "version": "1.13.1",
+  "version": "1.15.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/components/FilterTopPlanter.js
+++ b/src/components/FilterTopPlanter.js
@@ -75,6 +75,21 @@ function FilterTopPlanter(props) {
     props.onSubmit && props.onSubmit(filter);
   }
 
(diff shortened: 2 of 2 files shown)


thejoshcooter:
Features added for the issue, but I did notice that on the Verify view we use a form setup whereas we do not on the Planter view. These two could also probably be refactored into a single component later to ensure they stay in sync as things change which was one of the primary reasons behind this ticket.

nmcharlton:
> Features added for the issue, but I did notice that on the Verify view we use a form setup whereas we do not on the Planter view. These two could also probably be refactored into a single component later to ensure they stay in sync as things change which was one of the primary reasons behind this ticket.

That's a great shout. We use filters like this in a few places and will be adding more soon. A reusable component that allows us to configure the fields and specify the Apply handler would be a great help.

nmcharlton (approved):
Excellent work!

Issue #130, opened 2021-07-10
Edit Planter: Don't show blank selfies


nmcharlton:
To populate the selfie gallery in the _Edit Planter_ dialog we check the `planterPhotoUrl` values for all of the captures by the selected planter.
Currently, if any of these values URLs are blank (empty string), we show a blank image.

![Screenshot 2021-07-10 at 15 39 56](https://user-images.githubusercontent.com/5558838/125166791-15f9c800-e195-11eb-830d-fdb5da3b05e4.png)

We should filter out blank URLs.

To get to the _Edit Planter_ dialog, go to the _Planters_ tool, select a planter and click the edit icon.

nmcharlton:
Fixed under #139

PR #139, opened 2021-07-26
Edit Planter: Don't show blank selfies #130

brendongvieira:
Added  a line of code to filter out blank URLs, so we don't show the blank images anymore.


nmcharlton (approved):
Nice work, thanks @brendongvieira

PR #140, opened 2021-07-28
fix: date selection in filters

### LINKED ISSUE #94: Filter date selection: Cannot move forward

nmcharlton:
Resolves #94 

A `null` date passed into `KeyboardDatePicker` as `minDate` or `maxDate` causes the back and forward buttons respectively to be permanently disabled.

`1900-01-01` and `2100-01-01` are the default values for these properties (see https://material-ui-pickers.dev/api/DatePicker).

### DIFF
--- a/src/components/Filter.js
+++ b/src/components/Filter.js
@@ -270,7 +270,7 @@ function Filter(props) {
             value={dateStart}
             onChange={handleDateStartChange}
             format={getDateFormatLocale(true)}
-            maxDate={dateEnd || Date()}
+            maxDate={dateEnd || Date()} // Don't allow selection after today
             KeyboardButtonProps={{
               'aria-label': 'change date',
             }}
@@ -284,7 +284,7 @@ function Filter(props) {
             onChange={handleDateEndChange}
             format={getDateFormatLocale(true)}
             minDate={dateStart || datePickerDefaultMinDate}
-            maxDate={Date()}
+            maxDate={Date()} // Don't allow selection after today
             KeyboardButtonProps={{
               'aria-label': 'change date',
             }}
--- a/src/components/FilterTop.js
+++ b/src/components/FilterTop.js
@@ -235,7 +235,7 @@ function Filter(props) {
                   format={getDateFormatLocale(true)}
                   value={dateStart}
                   onChange={handleDateStartChange}
-                  maxDate={dateEnd || Date()}
+                  maxDate={dateEnd || Date()} // Don't allow selection after today
                   KeyboardButtonProps={{
                     'aria-label': 'change date',
                   }}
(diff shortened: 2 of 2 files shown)


src/components/Filter.js
```diff
@@ -285,7 +285,7 @@ function Filter(props) {
             value={dateStart}
             onChange={handleDateStartChange}
             format={getDateFormatLocale(true)}
-            maxDate={dateEnd}
+            maxDate={dateEnd || Date('2100-01-01')}
```
gwynndp, line 288: Perhaps we could add these default values to the state on line 58-59 so we can keep the initial state and it's defaults in one place?

src/components/FilterTop.js
```diff
@@ -230,7 +230,7 @@ function Filter(props) {
                   format={getDateFormatLocale(true)}
                   value={dateStart}
                   onChange={handleDateStartChange}
-                  maxDate={dateEnd}
+                  maxDate={dateEnd || Date('2100-01-01')}
```
gwynndp, line 233: Same as with Filter.js we can set defaults on lines 72-73 so they are set from the start.

src/components/Filter.js
nmcharlton: If I've understood your suggestion correctly, that doesn't have quite the desired result.
We want to keep the actual `dateStart` and `dateEnd` unset so that no value is selected in the input.
It's only the `minDate` and `maxDate` props that we need to set.

gwynndp: Ah, I misunderstood!  Ok, that solution works.  What do you think about adding comments to explain why the dates are there? or to create variables maxDateDefault and minDateDefault?  I'm concerned with someone changing or deleting the defaults in a refactor.
This would be a good thing to have a test for...

nmcharlton: That's a good shout, thanks. I'll make it a bit more explicit and add a test or two.

dadiorchen:
@nikolamilijevic1 what's your opinion in terms of the usage of the cypress unit test, is it okay? Or there still are some barriers using it.

nmcharlton:
The "fix" doesn't work quite as it should; needs a little more work 🤔

nmcharlton:
Modified behaviour to not allow date selection after _today_ (since all data in the platform is historic), but to allow arbitrary start and end date selection in the past if the other value has not been set.

PR #141, opened 2021-08-03
feat: add device identifier to planter detail dialog #36

### LINKED ISSUE #36: Add Device Identifier to Planter Detail dialog
EDIT: Filtering by Device Identifier now captured in #116.

~~It would be helpful to be able to filter Planters by their device identifier~~ (part of the `planter_registration` record linked to the planter). Note that there may be multiple `planter_registration` records, and therefore device identifiers associated with a planter.

Screenshot 2021-02-28 at 22 30 26

We should also add this field to the Planter Detail dialog.

Screenshot 2021-02-28 at 22 31 35

The corresponding API issue is Greenstand/treetracker-admin-api#532

chromium-52:
Resolves #36


nmcharlton (approved):
Beautifully done 😃

PR #142, opened 2021-08-04
feat: planter selfie rotation

### LINKED ISSUE #137: Planter selfie rotation
There is a legacy problem that's likely to never go away whereby planter images come into the platform with no EXIF rotation data. This means they need to be subsequently rotated for display, a process that's currently done using a slow, manual process that's not particularly scaleable.

The proposed workaround is to allow Admin Panel users with the Admin or Planter Manager role to rotate images (by increments of 90 degrees) and then storing the rotation value against the planter or image.
The rotation value will then be checked when rendering the photo in the _Planters_ tool and _Planter Detail_ dialog.

This rotation option should be available in the _Edit Planter_ dialog.

tranquanghuy0801:
Resolves #137

### DIFF
--- a/src/components/ImageScroller.js
+++ b/src/components/ImageScroller.js
@@ -19,7 +19,7 @@ const useStyle = makeStyles((theme) => ({
     width: `${IMAGE_CARD_SIZE * NUM_IMAGE_CARDS}px`,
   },
   imageList: {
-    height: `${IMAGE_CARD_SIZE + 16}px`,
+    height: `${IMAGE_CARD_SIZE}px`,
     display: 'flex',
     flexDirection: 'column',
     flexWrap: 'wrap',
@@ -41,7 +41,7 @@ const useStyle = makeStyles((theme) => ({
     height: '100%',
   },
   selectedImageCard: {
-    border: `solid ${theme.spacing(1.5)}px ${theme.palette.primary.main}`,
+    border: `inset ${theme.spacing(1.5)}px ${theme.palette.primary.main}`,
     margin: 0,
   },
   scrollButton: {
@@ -123,7 +123,7 @@ export default function ImageScroller(props) {
   }
 
   function handleImageChange(img) {
-    onSelectChange('imageUrl', img)
+    onSelectChange('imageUrl', img);
     if (images.length > 1) {
       setRotation(0);
     }
@@ -150,8 +150,8 @@ export default function ImageScroller(props) {
             >
               <OptimizedImage
                 src={img}
-                width={182}
-                height={192}
+                width={IMAGE_CARD_SIZE}
+                height={IMAGE_CARD_SIZE}
                 className={classes.image}
                 fixed
                 rotation={img === selectedImage ? rotation : 0}
--- a/src/components/PlanterDetail.js
+++ b/src/components/PlanterDetail.js
@@ -20,12 +20,14 @@ import EditPlanter from './EditPlanter';
 import OptimizedImage from './OptimizedImage';
 import LinkToWebmap from './common/LinkToWebmap';
 
+const PLANTER_IMAGE_SIZE = 441;
+
 const useStyle = makeStyles((theme) => ({
   box: {
     padding: theme.spacing(4),
   },
   cardMedia: {
-    height: '378px',
+    height: `${PLANTER_IMAGE_SIZE}px`,
   },
   personBox: {
     display: 'flex',
@@ -50,7 +52,7 @@ const useStyle = makeStyles((theme) => ({
   },
   imageContainer: {
     position: 'relative',
-    height: '378px',
+    height: `${PLANTER_IMAGE_SIZE}px`,
   },
 }));
 
@@ -107,10 +109,7 @@ const PlanterDetail = (props) => {
       <Drawer anchor="right" open={props.open} onClose={props.onClose}>
         <Grid
           style={{
-            width:
-              planter.imageRotation === 0 || planter.imageRotation === 180
-                ? 441
-                : 378,
+            width: PLANTER_IMAGE_SIZE,
           }}
         >
           <Grid container direction="column">
@@ -134,8 +133,8 @@ const PlanterDetail = (props) => {
               {planter.imageUrl && (
                 <OptimizedImage
                   src={planter.imageUrl}
-                  width={441}
-                  height={378}
+                  width={PLANTER_IMAGE_SIZE}
+                  height={PLANTER_IMAGE_SIZE}
                   className={classes.cardMedia}
                   fixed
                   rotation={planter.imageRotation}
--- a/src/components/Planters.js
+++ b/src/components/Planters.js
@@ -26,6 +26,8 @@ import LinkToWebmap from './common/LinkToWebmap';
 
 const log = require('loglevel').getLogger('../components/Planters');
 
+const PLANTER_IMAGE_SIZE = 182;
+
 const useStyles = makeStyles((theme) => ({
   outer: {
     height: '100vh',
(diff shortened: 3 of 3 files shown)


src/components/ImageScroller.js
```diff
@@ -129,7 +165,15 @@ export default function ImageScroller(props) {
                 height={192}
                 className={classes.image}
                 fixed
+                style={{ transform: `rotate(${rotation}deg)` }}
               />
+              <IconButton
```
nmcharlton, line 170: We should only show the rotation button on (and apply to the rotation to) the selected image.
Otherwise it looks a bit messy with multiple images:

https://user-images.githubusercontent.com/5558838/128629855-3437f0e5-b1e6-40f3-bf75-1d3d36c258c6.mov

src/components/ImageScroller.js
```diff
@@ -103,6 +126,19 @@ export default function ImageScroller(props) {
     });
   }
 
+  function handleDataChange(img) {
+    if (img) {
```
nmcharlton, line 130: Doing the selection and rotation at the same time doesn't work. Only the rotation is applied.
Selection and rotation should be two separate actions.

src/components/EditPlanter.js
```diff
@@ -90,17 +90,13 @@ const EditPlanter = (props) => {
       return newPlanter[key] !== planter[key];
     });
 
-    if (changed) {
+    if (key === 'imageRotation' || changed) {
```
nmcharlton, line 93: This causes the _Save_ button to be enabled if there is a non-zero rotation, even if nothing has changed.

src/components/ImageScroller.js
```diff
@@ -65,20 +73,35 @@ const useStyle = makeStyles((theme) => ({
...
   },
+  clickRotate: {
+    margin: 5,
+    position: 'absolute',
+    bottom: 0,
+    right: 0,
+    width: 25,
+    height: 25,
+    color: 'white',
+    backgroundColor: 'black',
```
nmcharlton, line 84: If you use a `Fab` component, there's no need to override the default colors.

src/components/ImageScroller.js
```diff
@@ -129,7 +165,15 @@ export default function ImageScroller(props) {
                 height={192}
                 className={classes.image}
                 fixed
+                style={{ transform: `rotate(${rotation}deg)` }}
               />
+              <IconButton
+                className={classes.clickRotate}
+                onClick={() => handleDataChange(null)}
+                style={{ transform: `rotateY(180deg)` }}
```
nmcharlton, line 173: It's just the icon we want to rotate, not the button.

nmcharlton (changes requested):
It's a good start, but a few things don't quite work as they should.
We also need to apply the rotation on _Planter_ cards and in the _Planter Detail_ dialog.

nmcharlton:
@tranquanghuy0801 For info, while having a look at the latest updates I noticed a bug in the api (fixed in Greenstand/treetracker-admin-api#587) whereby `imageRotation` was required in a planter `PATCH`.

The fix has now been deployed to dev and test.

tranquanghuy0801:
@nmcharlton Yeah thanks for fixing that issue. I will check my code again later today.

tranquanghuy0801:
Screenshot 2021-08-10 at 19 58 03
@nmcharlton  Now I am getting stuck at this. As the Card image and the rotation icon both have the onClick props, therefore when I click the icon, it will also execute the onClick of the Card image, which will overwrite the planterUpdate variable

`planterUpdate = {imageRotation: 90} --> planterUpdate =  {imageUrl: ''}` 

This will make the "Save" button become inactive even you click rotate icon. Hope this make sense. If you do not understand, please let me know.

tranquanghuy0801:
@nmcharlton that's alright. I knew how to fix that. Check my latest commits.

nmcharlton (changes requested):
Thanks @tranquanghuy0801. This looks a lot better, but I still see a couple of issues:

1. The rotation should only be applied to the _selected_ image in the Edit Planter dialog, but is applied to all of them. Search planter ID 323 in the Planters tool for an example.
2. Rotated images have gaps at the top and bottom in the Edit Planter dialog, and overlap text in the Planter Detail drawer (see video below). The image should fill but not overflow the container.
3. If you edit the planter from the Verify tool, the Planter Detail does not get updated straight away. This is a more general problem, and not specific to image rotation, so could be handled under a separate PR if you don't want to fix it here.


https://user-images.githubusercontent.com/5558838/129447436-00b331ca-84d6-4f1e-93ee-96595cae9c85.mov

tranquanghuy0801:
@nmcharlton 
Thanks for your feedback. I fixed the issue 1 & 2. About the issue 3 you mentioned, you can create another issue for this and assign that to me.

src/components/ImageScroller.js
```diff
@@ -30,7 +30,7 @@ const useStyle = makeStyles((theme) => ({
     textAlign: 'center',
   },
   imageCard: {
-    height: '100%',
+    height: '84%',
```
nmcharlton, line 33: Let's keep it simple and make these images square.

src/components/ImageScroller.js
```diff
@@ -88,7 +88,16 @@ export default function ImageScroller(props) {
...
   const [maxImages, setMaxImages] = useState(MAX_IMAGES_INCREMENT);
   const imageScrollerRef = useRef(null);
-  let [rotation, setRotation] = useState(imageRotation);
+  let objRotation = {};
+  if (selectedImage) {
+    const selectedIndex = images.indexOf(selectedImage);
+    objRotation[selectedIndex] = imageRotation;
+  } else {
+    images.map((_, idx) => {
+      objRotation[idx] = imageRotation;
```
nmcharlton, line 97: It doesn't really make sense to apply image rotation to all images if none is selected.
Just leave all other images unrotated and apply the rotation to the selected image only.

src/components/ImageScroller.js
```diff
@@ -88,7 +88,16 @@ export default function ImageScroller(props) {
...
+    });
+  }
+  let [rotations, setRotations] = useState(objRotation);
```
nmcharlton, line 100: I think this is introducing unnecessary complexity.
There's no point in persisting the image rotation for other images in the frontend like this; only one value gets stored for each planter so all other rotations will be lost when they refresh (which is annoying for a user).

src/components/ImageScroller.js
```diff
@@ -147,14 +157,14 @@ export default function ImageScroller(props) {
                 height={192}
                 className={classes.image}
                 fixed
-                rotation={rotation}
+                rotation={rotations[idx] ? rotations[idx] : imageRotation}
```
nmcharlton, line 160: I'd suggest reverting back to the single value, but only applying it if this is the selected image.
When the user selects a new image, just reset the rotation to zero.

src/components/PlanterDetail.js
```diff
@@ -108,7 +105,14 @@ const PlanterDetail = (props) => {
   return (
     <React.Fragment>
       <Drawer anchor="right" open={props.open} onClose={props.onClose}>
-        <Grid className={classes.root}>
+        <Grid
```
nmcharlton, line 108: Let's make the container and image square to avoid this awkwardness and inconsistency between rotated and unrotated images.

The planter card can also be updated to use a square image.

nmcharlton (changes requested):
Sorry to keep asking for changes – this is a more complex problem than I'd realised!

tranquanghuy0801:
@nmcharlton 

I am a bit struggling to remove the white gaps when setting the rotation of image = 90 or 270 deg. Do you have any suggestions at this?

nmcharlton:
> @nmcharlton
> 
> I am a bit struggling to remove the white gaps when setting the rotation of image = 90 or 270 deg if I set the images square as you said in Edit Planter dialog. Do you have any suggestions at this?

@tranquanghuy0801 It seems to be a quirk of the way rotated elements are rendered, presumably due to some rounding in the transform calculation.
As a workaround I'd suggest increasing the height and width of the image slightly (just make sure you add a comment explaining the css. This appears to work in the `OptimizedImage` `style` property:
```
          width: 'calc(100% + 1px)',
          height: 'calc(100% + 1px)',
```

nmcharlton (approved):

PR #143, opened 2021-08-08
feat: improve consistency between Tree Detail and Planter Detail dialogs #10

### LINKED ISSUE #10: Improve consistency between Tree Detail and Planter Detail dialogs
Currently, planter details (and tree details on the Trees page) are shown on a side drawer, while tree details on the Verify page are shown as a modal popover dialog.

The two UIs could be more consistent.

Screenshot 2020-10-02 at 16 18 31

Screenshot 2020-10-02 at 16 16 53

chromium-52:
Resolves #10

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -76,25 +76,27 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 function CaptureDetailDialog(props) {
-  const { open, TransitionComponent, capture } = props;
+  const { open, capture } = props;
 
   const [snackbarOpen, setSnackbarOpen] = useState(false);
   const [snackbarLabel, setSnackbarLabel] = useState('');
   const [renderCapture, setRenderCapture] = useState(capture);
   const [screenWidth, setScreenWidth] = useState(window.innerWidth);
-  const resizeWidth = useCallback(() => {
+  const [screenHeight, setScreenHeight] = useState(window.innerHeight);
+  const resizeWindow = useCallback(() => {
     setScreenWidth(window.innerWidth);
+    setScreenHeight(window.innerHeight);
   }, []);
   const classes = useStyles();
 
   useEffect(() => {
     props.captureDetailDispatch.getCaptureDetail(props.capture.id);
 
-    window.addEventListener('resize', resizeWidth);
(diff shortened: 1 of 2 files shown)


src/components/CaptureDetailDialog.js
```diff
@@ -132,15 +156,26 @@ function CaptureDetailDialog(props) {
...
+            <Grid item>
+              <Box m={4}>
+                <Typography color="primary" variant="h6">
+                  Capture <LinkToWebmap value={capture.id} type="tree" />
+                  <CopyButton label="Capture ID" value={capture.id} />
+                </Typography>
+              </Box>
+            </Grid>
+            <Grid item>
+              <IconButton onClick={() => props.onClose()}>
```
nmcharlton, line 171: Clicking the _Close_ button throws an exception:
> Uncaught TypeError: props.onClose is not a function

src/components/CaptureDetailDialog.js
```diff
@@ -261,40 +296,42 @@ function CaptureDetailDialog(props) {
...
+        style={{ width: screenWidth - 340 }}
+        BackdropProps={{
+          classes: {
+            root: classes.dialog,
+          },
+        }}
+      >
+        <OptimizedImage
+          src={renderCapture.imageUrl}
+          width={320}
```
nmcharlton, line 313: I think we can make the capture image bigger now there's a bit more room for it.
We have three options, any of which would work well:

1. Dynamically calculate what size of image would fit in the space and request that using the `OptimizedImage` component
2. Drop `OptimizedImage` for the full-size image and just shrink the original image to fill the space
3. Set the `OptimizedImage` width to `600`

nmcharlton (changes requested):
Fantastic work! Just a couple of comments on the changes.

Some other suggestions for adjacent changes:
- I think we can get rid of the slide up transition – it looks a little odd when placed next to the sliding drawer.
- Can we also not clear the capture on `handleClose` in `Verify.js`? That will give a smooth when closing the dialog.

chromium-52:
Thank you so much for the comments! I've made the above changes; please let me know if I misunderstood anything or if I can improve further!

nmcharlton (approved):
This is awesome! Great work @chromium-52

PR #157, opened 2021-09-18
fix: replace logo png with svg

### LINKED ISSUE #150: Replace Greenstand logo png with svg or native text
The Greenstand logo with text that appears at the top of the menu on the home screen and in the toolbar on some pages is blurry due to scaling.

Screenshot 2021-08-14 at 17 18 06

It should be replaced with a crisp logo (either by rendering the png as an svg or by creating the Greenstand part with a native text element).

tranquanghuy0801:
Resolves #150


nmcharlton (approved):
Looks good

Issue #148, opened 2021-08-14
Rename "Planters" to "Growers"


nmcharlton:
The correct term for the people we currently refer to as _planters_ is _growers_.
This is because they do far more than plant trees; they maintain and support them as they grow, and our terminology should reflect that.

Similar to the recent renaming of _trees_ to _captures_, this should address all instances of the term _planter_, including in component, function and variable names and routes.

For now, the API will continue to use the term _planter_ until functionality is moved to the new microservices architecture under the domain migration.

One area that will require attention is the user roles, one of which is "Planter Manager".

PR #163, opened 2021-09-20
feat: refactoring renaming planters to growers

tranquanghuy0801:
resolve the issue #148


src/context/GrowerContext.js
```diff
@@ -0,0 +1,125 @@
+import React, { useState, createContext } from 'react';
```
nmcharlton, line 1: git has picked up this renaming as a deletion and creation of a new file.
Are you able to redo this file using a `git mv` command so we keep the history and can compare?

src/common/policy.js
```diff
@@ -17,12 +17,12 @@ const policy = {
       description: 'Can approve/reject trees',
     },
     {
-      name: 'list_planter',
-      description: 'Can view planters',
+      name: 'list_grower',
```
nmcharlton, line 20: Unfortunately, this policy name value is encoded in records in the `admin_role` table in the database.
We should aim to eventually change it, but it will require some coordination across the deployments.

src/models/auth.js
```diff
@@ -10,7 +10,7 @@ const POLICIES = {
   MANAGER_USER: 'manager_user',
   LIST_TREE: 'list_tree',
   APPROVE_TREE: 'approve_tree',
-  LIST_PLANTER: 'list_planter',
+  LIST_GROWER: 'list_GROWER',
```
nmcharlton, line 13: The key could change to `LIST_GROWER` but the value will need to remain as `list_planter` for compatibility with the production database.

nmcharlton (changes requested):

tranquanghuy0801:
@nmcharlton , I have changed the filename of all relevant to the original ones as you requested.

nmcharlton:
> @nmcharlton , I have changed the filename of all relevant to the original ones as you requested.

Sorry Harry, that's not what I meant. There was one file where we'd lost the history during the renaming, but all the other files were fine.

tranquanghuy0801:
> > @nmcharlton , I have changed the filename of all relevant to the original ones as you requested.
> 
> Sorry Harry, that's not what I meant. There was one file where we'd lost the history during the renaming, but all the other files were fine.

Sorry Nick, I have changed back to what I did initially. The only two files I keep the original names are src/components/PlanterContext.js and cypress/components/planter.spec.py.js as git recognized these files as being new.

nmcharlton (approved):

Issue #159, opened 2021-09-19
Filter by organisation Name in planter tool


sebastiangaertner:
See screenshot. Currently you have to know the organization ID which is not easy. This would be a select by "o" or "O" in entity.type
![Screenshot 2021-09-19 at 15 08 38](https://user-images.githubusercontent.com/3283800/133926982-6da084b5-5470-4d9b-90c2-331c3aacb7fc.png)

Issue #161, opened 2021-09-20
Critical: Can not filter by Organization id when logged in with user role


sebastiangaertner:
See attached. This is a user under freetown policy in "public.admin_role" linking organization_id = 178  "public.entity_relationship" with the organizations_ids 838, 183, 839, 184, 179, 840, 836, 841, 842, 837, 182.
When logging in under a Freetown user and filtering by organization_id the filter does not work. see image attached. 
![Screenshot 2021-09-20 at 09 37 01](https://user-images.githubusercontent.com/3283800/133964788-b765c1b6-cf8b-4cc9-a084-30d623af8579.png)

Mloweedgar:
I would like to work  on this Issue. Since I'm kind new to the code-base can you help me fill the template below so that I can be able to reproduce and fix the bug :point_down: 

## Description


## How to reproduce

-  
- 

*Expected Result*


*Current Result*



- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>


(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

PR #164, opened 2021-09-20
fix: planter filter to allow filtering parent orgs by name and all orgs by id

gwynndp:
- replaced text box with an org dropdown to filter orgs on Planters page
- admin can filter by parent org names in a dropdown on both Planters and Verify pages
- org account can filter by sub-org names in a dropdown on both Planters and Verify pages

There is a related PR for the api here: https://github.com/Greenstand/treetracker-admin-api/pull/589

Closes: https://github.com/Greenstand/treetracker-admin-client/issues/159, https://github.com/Greenstand/treetracker-admin-client/issues/161


nmcharlton (approved):
Nice work!

PR #166, opened 2021-09-23
fix: add uuid to capture detail dialog and top filter

### LINKED ISSUE #110: Add UUID to Verify filter and Capture Detail
It should be possible to filter on a capture's UUID (`uuid` property).
Initially, this should be in the Verify tool, but #51 will align the Verify and Captures filters.
For consistency, the UUID should also be visible in the Capture Detail dialog (as a copyable item).

Currently there is no way to pull up a particular capture in the Admin Panel based on information in an image's file name (`TimeStamp-LatLong-UUID-OrigImageName.jpg`).
Date can be used, but this may pull up thousands of images.

tranquanghuy0801:
Resolves #110

### DIFF
--- a/src/components/FilterTop.js
+++ b/src/components/FilterTop.js
@@ -168,14 +168,6 @@ function Filter(props) {
           <Grid container wrap="nowrap" direction="row">
             <Grid item className={classes.inputContainer}>
               <TextField
-                htmlFor="uuid"
-                id="uuid"
-                label="UUID"
-                placeholder=""
-                value={uuid}
-                onChange={(e) => setUUID(e.target.value)}
-              />
-              <TextField
                 select
                 htmlFor="verification-status"
                 id="verification-status"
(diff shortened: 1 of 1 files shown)


nmcharlton (approved):

src/components/FilterTop.js
```diff
@@ -164,6 +167,14 @@ function Filter(props) {
         <form onSubmit={handleSubmit}>
           <Grid container wrap="nowrap" direction="row">
             <Grid item className={classes.inputContainer}>
+              <TextField
```
nmcharlton, line 170: I'm not sure this field should go at the beginning of the filters.
Maybe after Capture ID would be better.

nmcharlton (changes requested):

src/components/FilterTop.js
tranquanghuy0801: @nmcharlton , I have updated as you requested

nmcharlton (approved):

PR #168, opened 2021-09-30
feat: adding Export Captures button 

### LINKED ISSUE #52: Export Captures as csv
The Trees page needs an Export button that exports all trees included in the current filter to a csv file.
Pressing Export should generate a popup dialog that allows the user to select/deselect the columns they'd like to include (all selected by default).
When the user clicks Export in the popup dialog, the browser initiates a download of the generated csv file (including column headers).

Note that all filter matches should be included, not just those visible on the current page.

tranquanghuy0801:
Resolves #52

### DIFF
--- a/src/context/CapturesContext.js
+++ b/src/context/CapturesContext.js
@@ -137,11 +137,7 @@ export function CapturesProvider(props) {
     console.log('captures filterInfo -- ', filterInfo);
 
     // if filterInfo contains new values override the defaults in state hooks
-    const {
-      filter = new FilterModel(),
-      orderBy = 'id',
-      order = 'asc',
-    } = filterInfo;
+    const { filter = new FilterModel() } = filterInfo;
 
     const where = filter ? filter.getWhereObj() : {};
 
@@ -166,7 +162,7 @@ export function CapturesProvider(props) {
     const paramString = `filter=${JSON.stringify(lbFilter)}`;
     const response = await queryCapturesApi({ paramString });
     return response;
-  }
+  };
 
   const getCaptureAsync = (id) => {
     queryCapturesApi({ id })
--- a/src/components/ExportCaptures.js
+++ b/src/components/ExportCaptures.js
@@ -10,9 +10,10 @@ import {
   FormControl,
   FormGroup,
   FormControlLabel,
+  DialogContentText,
 } from '@material-ui/core';
 import { CapturesContext } from '../context/CapturesContext';
-import { CSVLink } from "react-csv";
+import { CSVLink } from 'react-csv';
 import { formatCell } from './Captures/CaptureTable';
 import { LinearProgress } from '@material-ui/core';
 
@@ -26,7 +27,7 @@ const useStyle = makeStyles((theme) => ({
     flexGrow: 1,
   },
   formControl: {
-    margin: theme.spacing(3),
+    margin: theme.spacing(2),
   },
 }));
 
@@ -35,7 +36,7 @@ const ExportCaptures = (props) => {
   const classes = useStyle();
   let nameColumns = {};
   columns.forEach(({ attr, renderer }) => {
-    nameColumns[attr] = {status: true, renderer: renderer};
+    nameColumns[attr] = { status: true, renderer: renderer };
   });
   // checkboxes to choose download columns
   const [checkedColumns, setCheckColumns] = useState(nameColumns);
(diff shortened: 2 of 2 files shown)


tranquanghuy0801:
Screenshot 2021-09-30 at 22 27 29

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+  });
+  const [checkedColumns, setCheckColumns] = useState(nameColumns);
+  const capturesContext = useContext(CapturesContext);
+  const [downloadData, setDownloadData] = useState([]);
+
+  function handleChange(attr) {
+    setCheckColumns({ ...checkedColumns, [attr]: false });
+  }
+
+  function processDownloadData(captures, selectedColumns) {
```
nmcharlton, line 53: There's a lot of unnecessary duplication in this function that will make it more difficult to maintain and error prone:

- Use the `renderer` values that are already defined in `columns` rather than defining them all again.
- If you must define new renderers for any reason, make that the only thing you set inside the `if - else` statement. Only include the `formatCapture[attr] = formatCell(...` statement once.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+    return captures.map(capture => {
+      let formatCapture = {};
+      Object.keys(selectedColumns).forEach(attr => {
+        if (attr === 'id' || attr === 'planterId') {
+          formatCapture[attr] = capture[attr];
+        } else if (attr === 'tokenId') {
+          formatCapture[attr] = formatCell(capture, speciesState, attr, (val) =>
```
nmcharlton, line 60: I'm don't think calling `formatCell` directly here is a good idea. If someone adds another `LinkToWebmap` field, they would need to update this function to avoid breaking CSV export.

Either split out `formatCell` into the content formatting and the presentation, or pass a flag to the function to request a plain text response.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+      <DialogActions>
+        <Button onClick={handleClose}>Cancel</Button>
+        <Button color="primary">
+          <CSVLink
+            className={classes.csvLink}
+            data={downloadData}
+            filename={"captures.csv"}
+            asyncOnClick={true}
+            onClick={() => {
+              let filterColumns = Object.entries(checkedColumns).filter((val) => val[1] === true);
```
nmcharlton, line 107: This can be `const`

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+        </FormControl>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={handleClose}>Cancel</Button>
+        <Button color="primary">
+          <CSVLink
+            className={classes.csvLink}
+            data={downloadData}
+            filename={"captures.csv"}
+            asyncOnClick={true}
```
nmcharlton, line 105: The `async` functionality isn't working for me (Chrome on MacOS).
A (nearly) empty CSV starts downloading immediately when I click the button.
The next time I click _Download CSV_ I get the data I expect.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+  let nameColumns = {};
+  columns.forEach(({ attr }) => {
+    nameColumns[attr] = true;
+  });
+  const [checkedColumns, setCheckColumns] = useState(nameColumns);
+  const capturesContext = useContext(CapturesContext);
+  const [downloadData, setDownloadData] = useState([]);
+
+  function handleChange(attr) {
+    setCheckColumns({ ...checkedColumns, [attr]: false });
```
nmcharlton, line 50: This should toggle the value, not always set it to false.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+          <CSVLink
+            className={classes.csvLink}
+            data={downloadData}
+            filename={"captures.csv"}
+            asyncOnClick={true}
+            onClick={() => {
+              let filterColumns = Object.entries(checkedColumns).filter((val) => val[1] === true);
+              const selectedColumns = Object.fromEntries(filterColumns);
+              capturesContext.getAllCaptures({ filter }).then((response) => {
+                setDownloadData(processDownloadData(response.data, selectedColumns));
```
nmcharlton, line 110: `processDownloadData` takes a long time and locks up the browser for a few seconds on a fairly powerful machine. On a slower machine this could be a lot longer.

As a minimum, we need to give the user visual feedback that the export is in progress, but even better would be to do the processing in a web worker.
If it's helpful, we can ask the Design team to create a visual workflow.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+              });
```
nmcharlton, line 111: The _Export Captures_ dialog should close when the user clicks `Download CSV`, and be replaced by the visual feedback on progress mentioned in the comment above.

src/components/ExportCapture.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  return (
+    <Dialog
+      open={isOpen}
+      onClose={handleClose}
+      aria-labelledby="form-dialog-title"
+    >
+      <DialogTitle id="form-dialog-title">Export Captures</DialogTitle>
+      <DialogContent>
```
nmcharlton, line 79: Add a label with of the number of captures to be downloaded. This would be a good place to warn the user that only the first 30,000 will be downloaded.

nmcharlton (changes requested):

tranquanghuy0801:
@nmcharlton 
Sorry for the delay on doing this. I was busy on other tasks at work.
For this issue, I have made some changes:

- Remove all conditions for other columns in function `processDownloadData()` except for the `id` and `planterID` as they are applied `<LinkToWebMap>` in formatCell
- Added a linear progress feedback at the end of Export Captures form when you click Download. I just put here for start, I think you might want to change this. You can ask the Design team for this feature and I will update then.
- Fixed the issue of only getting the data at 2nd download time.

src/context/CapturesContext.js
```diff
@@ -131,6 +132,42 @@ export function CapturesProvider(props) {
...
   };
 
+  const getAllCaptures = async (filterInfo = {}) => {
+    log.debug('load all captures');
+    console.log('captures filterInfo -- ', filterInfo);
+
+    // if filterInfo contains new values override the defaults in state hooks
+    const {
+      filter = new FilterModel(),
+      orderBy = 'id',
```
nmcharlton, line 142: We should use the same order as the table: `order: [`${orderBy} ${order}`]`
This is particularly important when the user hits their 20,000 limit.

src/context/CapturesContext.js
```diff
@@ -131,6 +132,42 @@ export function CapturesProvider(props) {
...
+      order = 'asc',
+    } = filterInfo;
+
+    const where = filter ? filter.getWhereObj() : {};
+
+    const lbFilter = {
+      where: { ...where },
+      order: [`${orderBy} ${order}`],
+      limit: 20000,
```
nmcharlton, line 151: We need to warn the user that only the first 20,000 records will be downloaded.

nmcharlton:
Well done, this is a lot better!
I think the progress bar in the dialog works well.

One final suggestions: Disable the buttons in the dialog while downloading is in progress and close the dialog when downloading is complete.

nmcharlton (changes requested):

nmcharlton (approved):
Excellent work @tranquanghuy0801, thank you!

PR #173, opened 2021-10-14
feat: include badge showing applied filters in Verify page

### LINKED ISSUE #81: Show which Verify filters are applied when filter dialog is hidden
In the Verify page, it's possible to apply a set of filters and hide the dialog.

At this point, the user has to remember what filters they selected, and it's easy to forget that you have applied a filter, especially when moving between tools in the Admin Panel.

This is particularly troublesome because the filter dialog is hidden by default when going back to the Verify page (see video).

https://user-images.githubusercontent.com/5558838/117187148-83732c00-add3-11eb-88a4-59c5b006f739.mov

tranquanghuy0801:
Resolves #81

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -9,35 +9,3 @@ export const getVerificationStatus = (active, approved) => {
     return verificationStates.REJECTED;
   }
 };
-
-export const countAppliedFilters = (filter) => {
-  let numFilters = 0;
-
-  if (!filter) {
-    return numFilters;
-  }
-
-  if (filter['active'] !== undefined && filter['approved'] !== undefined) {
-    numFilters += 1;
-  }
-
-  Object.keys(filter).forEach(key => {
-    if (key === 'filter' || key == 'active' || key == 'approved') {
-      return;
-    } else if ((key === 'captureId' || key === 'uuid' || key === 'deviceIdentifier' || key === 'planterId' || key === 'planterIdentifier') && filter[key] !== '') {
-      numFilters += 1;
-    } else if (filter['tagId'] > 0) {
-      numFilters += 1;
-    } else if ((key === 'dateStart' || key === 'dateEnd') && filter[key] !== undefined) {
-      numFilters += 1;
-    } else if (key == 'organizationId' && filter[key] !== 'ALL_ORGANIZATIONS') {
-      numFilters += 1;
-    } else if (key == 'speciesId' && filter[key] !== 'ALL_SPECIES') {
-      numFilters += 1;
-    } else if (key == 'tokenId' && filter[key] !== 'All') {
-      numFilters += 1;
-    }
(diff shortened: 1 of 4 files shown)


src/common/utils.js
```diff
@@ -8,4 +8,36 @@ export const getVerificationStatus = (active, approved) => {
   } else if (active === false && approved === false) {
     return verificationStates.REJECTED;
   }
+};
+
+export const countAppliedFilters = (filter) => {
```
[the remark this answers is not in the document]
nmcharlton: It would also be good to create a test for this function.

src/components/Verify.js
```diff
@@ -446,13 +454,22 @@ const Verify = (props) => {
...
                   onClick={handleFilterClick}
                   startIcon={<IconFilter />}
                   key={1}
                 >
                   Filter
+                  {
+                    countAppliedFilters(verifyContext.filter) > 0 ?
+                      <Avatar className={classes.activeFilters}>
+                      {
+                        countAppliedFilters(verifyContext.filter)
```
nmcharlton, line 468: There's no need to call the function twice.
Assign the result to a variable and check that instead.

src/components/Verify.js
```diff
@@ -446,13 +454,22 @@ const Verify = (props) => {
...
-                  variant="text"
+                  variant="outlined"
                   color="primary"
                   onClick={handleFilterClick}
                   startIcon={<IconFilter />}
                   key={1}
                 >
                   Filter
+                  {
+                    countAppliedFilters(verifyContext.filter) > 0 ?
```
nmcharlton, line 465: This can be expressed as:
```
(countAppliedFilters(verifyContext.filter) > 0) && <Avatar...
```
There's no need for the ternary operator.

src/common/utils.js
```diff
@@ -8,4 +8,36 @@ export const getVerificationStatus = (active, approved) => {
...
+  Object.keys(filter).forEach(key => {
+    if (key === 'filter' || key == 'active' || key == 'approved') {
+      return;
+    } else if ((key === 'captureId' || key === 'uuid' || key === 'deviceIdentifier' || key === 'planterId' || key === 'planterIdentifier') && filter[key] !== '') {
+      numFilters += 1;
+    } else if (filter['tagId'] > 0) {
+      numFilters += 1;
+    } else if ((key === 'dateStart' || key === 'dateEnd') && filter[key] !== undefined) {
+      numFilters += 1;
+    } else if (key == 'organizationId' && filter[key] !== 'ALL_ORGANIZATIONS') {
```
nmcharlton, line 33: We have `ALL_ORGANIZATIONS` and `ALL_SPECIES` defined as variables in `src/models/filter.js`. It's best to use those in place of these string literals.

PR #182, opened 2021-10-28
fix(capture): enables scrolling in Capture page

### LINKED ISSUE #176: Cannot scroll Captures page
Present in v1.22.1
The Captures page does not have a vertical scrollbar, so users cannot scroll content:

mqnguyen5:
Fixes #176 

This PR enables vertical scrolling in the Capture page by changing the `overflow: hidden` property into `overflowX: 'hidden'` and `overflowY:'scroll'`

osd_release_greenstand

### DIFF
--- a/src/components/CaptureTable.js
+++ b/src/components/CaptureTable.js
@@ -113,8 +113,6 @@ const CaptureTable = () => {
   const scrollRef = createRef();
   const classes = useStyle();
 
-  const [page, setPage] = useState(capturesContext.page);
-
   useEffect(() => {
     loadSpecies();
     loadCaptures();
@@ -159,9 +157,8 @@ const CaptureTable = () => {
   };
 
   const handlePageChange = (e, newPage) => {
-    setPage(newPage);
     loadCaptures({
-      page: page,
+      page: newPage,
       rowsPerPage: capturesContext.rowsPerPage,
       filter: capturesContext.filter,
     });
@@ -196,7 +193,7 @@ const CaptureTable = () => {
         rowsPerPageOptions={[25, 50, 100, 250, 500]}
         component="div"
         count={capturesContext.captureCount}
-        page={page}
+        page={capturesContext.page}
         rowsPerPage={capturesContext.rowsPerPage}
         onChangePage={handlePageChange}
         onChangeRowsPerPage={handleRowsPerPageChange}


src/components/CaptureTable.js
```diff
@@ -113,6 +113,8 @@ const CaptureTable = () => {
   const scrollRef = createRef();
   const classes = useStyle();
 
+  const [page, setPage] = useState(capturesContext.page);
```
nmcharlton, line 116: There's no need to store the `page` as local state – `CapturesContext` maintains this value for us, and we want to avoid duplicating it.
Were you trying to address the lag between clicking `>` and the page number changing, and that clicking multiple times doesn't result in the same number of page changes?
If so, I'd suggest looking at `CapturesContext:getCapturesAsync()` and moving the state updates before the API call:
```
    setPage(page);
    setRowsPerPage(rowsPerPage);
    setOrderBy(orderBy);
    setOrder(order);
    setFilter(filter);
```

nmcharlton (changes requested):

src/components/CaptureTable.js
mqnguyen5: Previously, the `handlePageChange()` function looks like this:

```js
const handlePageChange = () => {
    loadCaptures({
      page: capturesContext.page + 1, // the location where the bug happens
      rowsPerPage: capturesContext.rowsPerPage,
      filter: capturesContext.filter,
    });
  };
```

This caused the page to moving forward even when clicking `<`, while clicking `<` supposed to move us back to the previous page. When checking the [TablePagination API](https://mui.com/api/table-pagination/#props), I notice that `handlePageChange()` accepts 2 arguments: a React mouse event and the new page value. As such, I thought that I need to somehow change the `page` state into the new page value before calling `loadCaptures()`, which explains why I create a local `page` state. I'll try to find a way to make use of the `CapturesContext:getCapturesAsync()` per your suggestion, but I'm not sure how I can do it.

nmcharlton: You've fixed the problem well, you just don't need to set a local state for `page`. Just use `capturesContext.page` directly.

When you pass `page: newPage` in `loadCaptures()`, `capturesContext.page` gets updated.

mqnguyen5:
Hi @nmcharlton , I have updated the code, please have a look when you have time. Thanks.

nmcharlton (approved):
Excellent work

mqnguyen5:
Hi @nmcharlton , if possible, could you please add a `hacktoberfest-accepted` label to this PR? I would really appreciate it. Thanks!

PR #184, opened 2021-10-29
refactor: clean up CapturesTable

gwynndp:
## Description
Break out sub-components and styles and organize the files for Captures into a folder

**What kind of change does this PR introduce?** 
 - [x] Refactor components

## Issue

There is one big component CaptureTable that contains everything.

**What is the new behavior?**

Refactor captures components to separate the Filter from the Table view.  Will make testing and state management easier and reduce requests and rendering, making the app faster to respond to changes.

Steps taken:
 - [x]  Create new component FilterHeader that nests the FilterTop in the Nav
 - [x]  Move api request for captures to CapturesContext
 - [x]  Update state functions on CapturesContext to use the state directly
 - [x]  Remove unnecessary Grids around the CaptureTable
 - [x]  Move styles from CaptureTable to a file of their own, CaptureTable.styles
 - [x]  Move CaptureTable, CaptureTable.styles, and captures.test to Captures folder
 - [x]  Create a fixtures file to store sample data for tests
 - [x]  Update tests for CaptureTable 
 - [x]  Copy tests for FilterTop from organization.test
 - [x]  Fix scrolling and pagination on CaptureTable
 - [x]  Add filter count badge to CaptureFilterHeader

## Breaking change
 - [x] No

## No change to the display 
Capture Data - Treetracker Admin by Greenstand 2021-10-28 19-01-32

### DIFF
--- a/src/components/Captures/CaptureTable.js
+++ b/src/components/Captures/CaptureTable.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useContext, createRef } from 'react';
+import React, { useEffect, useState, useContext } from 'react';
 import {
   Grid,
   Table,
@@ -80,7 +80,6 @@ const CaptureTable = () => {
   const speciesContext = useContext(SpeciesContext);
   const [isDetailsPaneOpen, setIsDetailsPaneOpen] = useState(false);
   const [speciesState, setSpeciesState] = useState({});
-  const scrollRef = createRef();
   const classes = useStyle();
 
   useEffect(() => {
@@ -120,10 +119,10 @@ const CaptureTable = () => {
 
   const createSortHandler = (attr) => {
     return () => {
-      const order = orderBy === attr && order === 'asc' ? 'desc' : 'asc';
-      const orderBy = attr;
-      setOrder(order);
-      setOrderBy(orderBy);
+      const newOrder = orderBy === attr && order === 'asc' ? 'desc' : 'asc';
+      const newOrderBy = attr;
+      setOrder(newOrder);
+      setOrderBy(newOrderBy);
     };
   };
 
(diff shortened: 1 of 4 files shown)


nmcharlton (commented):
I haven't done a thorough review yet, but @tranquanghuy0801 is adding the Export Captures function within the Captures tool (#168). We'll need to make sure his changes don't get lost when we merge.

src/components/FilterHeader.js
```diff
@@ -0,0 +1,54 @@
...
+import { CapturesContext } from '../context/CapturesContext';
+import Navbar from '../components/Navbar';
+import FilterTop from '../components/FilterTop';
+
+function FilterHeader() {
+  const capturesContext = useContext(CapturesContext);
+  const [isFilterShown, setFilterShown] = useState(true);
+
+  useEffect(() => {
+    document.title = `Capture Data - ${documentTitle}`;
```
nmcharlton, line 14: I think this should be in `CapturesView`

src/components/FilterHeader.js
```diff
@@ -0,0 +1,54 @@
+import React, { useState, useContext, useEffect } from 'react';
```
nmcharlton, line 1: Can we rename this component to make it clear it's a Capture filter header?
We could rename `FilterTop` at the same time.

nmcharlton (approved):
A couple of suggestions, but nothing major.

src/components/FilterHeader.js
gwynndp: The useEffect for the documentTitle and the filter state? 
Or should I spread out the code for the Nav and FilterTop in the FilterHeader in the CaptureView instead of creating a new component?  Then I could rename FilterTop to CaptureFilter and we could use the same nesting for the Growers with a GrowersView and GrowersFilter?

nmcharlton: Just the doc title. We should be setting it in the view components, as they're the top level at which the title applies.
The additional refactoring you suggest does make sense, but could probably wait for now.

src/components/Captures/CaptureTable.js
```diff
@@ -0,0 +1,225 @@
...
+        rowsPerPage={rowsPerPage}
+        onChangePage={handlePageChange}
+        onChangeRowsPerPage={handleRowsPerPageChange}
+      />
+    );
+  };
+
+  return (
+    <Grid item container style={{ height: '100%', overflow: 'auto' }}>
+      <div className={classes.tableContainer} ref={scrollRef}>
```
nmcharlton, line 146: This container isn't filling the width of the window.

src/components/Captures/CaptureTable.js
```diff
@@ -0,0 +1,225 @@
...
+    setPage(page);
+  };
+
+  const handleRowsPerPageChange = (e) => {
+    setRowsPerPage(parseInt(e.target.value));
+  };
+
+  const createSortHandler = (attr) => {
+    return () => {
+      const order = orderBy === attr && order === 'asc' ? 'desc' : 'asc';
```
nmcharlton, line 123: Looks like we have a name clash on `order` and `orderBy` here (local vs props).

src/components/FilterHeader.js
nmcharlton, line 14: Can we remove this now?

src/components/tests/CaptureFilter.test.js
```diff
@@ -0,0 +1,182 @@
...
+      });
+
+      it('renders text "Verification Status" ', () => {
+        render(component);
+        expect(screen.getByText('Verification Status')).toBeInTheDocument();
+      });
+
+      it('renders "Start Date" input ', () => {
+        render(component);
+        let input = screen.getByRole('textbox', { name: 'Start Date' });
```
nmcharlton, line 53: I know it's _only_ a test script, but there are a lot of `let` statements in this file that could be `const`.

src/components/tests/CaptureFilter.test.js
```diff
@@ -0,0 +1,182 @@
...
+        const orglist = screen.getByRole('listbox');
+        const orgs = within(orglist).getAllByTestId('org-item');
+        const listItems = orgs.map((org) => org.textContent);
+        console.log('default orgList', listItems);
+
+        expect(orgs).toHaveLength(2);
+      });
+    });
+
+    // describe('context data renders in child', () => {
```
nmcharlton, line 151: Is this test awaiting implementation? Rather than comment it out, can it be either removed or marked `skip`?

src/views/CapturesView.js
```diff
@@ -1,17 +1,11 @@
...
-
-Captures is a container component (no visual representation of its own and concerned with
-handling the comms between the captures view components and the store/models)
-
-*/
 import React, { useEffect } from 'react';
 import { documentTitle } from '../common/variables';
 import { Grid } from '@material-ui/core';
-import CaptureTable from '../components/CaptureTable';
+import CaptureTable from '../components/Captures/CaptureTable.js';
```
nmcharlton, line 4: For consistency, we should omit the `.js` on the end.

nmcharlton (changes requested):

src/components/Captures/CaptureTable.js
gwynndp: Thanks for catching that!

src/components/Captures/CaptureTable.js
gwynndp: Sorry, I can't see where the clash is -- they should only be defined via context now, I removed props from the component.  Can you give me more info?

PR #195, opened 2021-11-13
chore: create issue templates

nmcharlton:


nmcharlton:
@gwynndp Can you see what you think of these (default) issue templates and suggest any improvements?

gwynndp:
> @gwynndp Can you see what you think of these (default) issue templates and suggest any improvements?

I like them both.  One addition to the templates would be a link to the page or "section" where they have either found the bug or think the feature should live (e.g. CapturesTable, or Captures section).  Maybe it could be included in the meta-data at the top.

Is the goal to make it easier to gather information about work that needs doing or to make defined work orders?

These seem like they are good for submitting information that may need to be reviewed by someone before being turned into actual issues for developers to pick up.  I do think we need something like this, but they may open more questions rather than defining an issue for a programmer to pick up and implement.  

Could we create a 3rd template along the same lines where the spec is defined?  This would be user-friendly for the fly-by volunteer coder:
- Describe goal feature/user story, link to design, name the section where it should be implemented, info on api (if needed)
- Include basic project info to get up and running
- Include instructions on how to do a PR, etc.



Resources:
- https://www.freecodecamp.org/news/how-to-attract-new-contributors-to-your-open-source-project-46f8b791d787/
- Because people are funny: https://www.talater.com/open-source-templates/#/
 I tried it out and got ........................
 
 

## Expected Behavior



## Current Behavior



## Possible Solution



## Steps to Reproduce (for bugs)


1.
2.
3.
4.

## Context



## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

nmcharlton:
Closing as we seem to be managing ok without templates

PR #200, opened 2021-11-16
Stakeholder tool

gwynndp:
DRAFT
- move state from redux to context
- add api functions to mock api
- fix opening the modal for child-orgs

### DIFF
--- a/src/components/Stakeholders/Filter.js
+++ b/src/components/Stakeholders/Filter.js
@@ -35,7 +35,7 @@ const useStyles = makeStyles({
   },
 });
 
-function Filter() {
+function StakeholderFilter() {
   const classes = useStyles();
   const { orgList, userHasOrg } = useContext(AppContext);
 
@@ -248,4 +248,4 @@ function Filter() {
   );
 }
 
-export default Filter;
+export default StakeholderFilter;
--- a/src/components/Stakeholders/FilterBar.js
+++ b/src/components/Stakeholders/FilterBar.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
-import Filter from './Filter';
+import StakeholderFilter from './Filter';
 import Add from './Add';
 
 const useStyles = makeStyles({
@@ -17,13 +17,13 @@ const useStyles = makeStyles({
   },
 });
 
-export default function FilterBar() {
+export default function StakeholderFilterBar() {
   const classes = useStyles();
 
   return (
     <div className={`${classes.flex} ${classes.spaceBetween} ${classes.my}`}>
       <div>
-        <Filter />
+        <StakeholderFilter />
       </div>
       <div className={classes.flex}>
         <Add />
--- a/src/components/Stakeholders/StakeholderDetail/Header.js
+++ b/src/components/Stakeholders/StakeholderDetail/Header.js
@@ -43,7 +43,7 @@ const useStyles = makeStyles({
   },
 });
 
-export default function DialogHeader({ data }) {
+export default function StakeholderDialogHeader({ data }) {
   const { updateStakeholder } = useContext(StakeholdersContext);
   const [isEditing, setIsEditing] = useState(false);
   const [details, setDetails] = useState({ ...data });
--- a/src/components/Stakeholders/Table.js
+++ b/src/components/Stakeholders/Table.js
@@ -81,7 +81,7 @@ function StakeholderTable() {
         <TablePagination
           component="div"
           rowsPerPage={rowsPerPage}
-          rowsPerPageOptions={[1, 2, 5]}
+          rowsPerPageOptions={[5, 10, 20]}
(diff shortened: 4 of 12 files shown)


src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,251 @@
...
+  flex: {
+    display: 'flex',
+    flex: 'wrap',
+  },
+  ml: {
+    marginLeft: '24px',
+  },
+});
+
+function Filter() {
```
nmcharlton, line 38: Can we call this component `StakeholderFilter`?

src/components/Stakeholders/FilterBar.js
```diff
@@ -0,0 +1,33 @@
...
+  spaceBetween: {
+    justifyContent: 'space-between',
+  },
+  my: {
+    marginTop: '3rem',
+    marginBottom: '2rem',
+  },
+});
+
+export default function FilterBar() {
```
nmcharlton, line 20: Can this be `StakeholderFilterBar`?

src/components/Stakeholders/StakeholderDetail/Header.js
```diff
@@ -0,0 +1,322 @@
...
+  },
+  inputName: {
+    fontSize: '1.67rem',
+  },
+  input: {
+    height: 31,
+  },
+});
+
+export default function DialogHeader({ data }) {
```
nmcharlton, line 46: Add _Stakeholder_ to the component name

src/components/Stakeholders/Table.js
```diff
@@ -0,0 +1,95 @@
...
+                    />
+                  ))}
+              </React.Fragment>
+            ))}
+          </TableBody>
+        </Table>
+        <TablePagination
+          component="div"
+          rowsPerPage={rowsPerPage}
+          rowsPerPageOptions={[1, 2, 5]}
```
nmcharlton, line 84: This doesn't seem right - is it just for testing?

src/context/AppContext.js
```diff
@@ -113,6 +115,18 @@ function getRoutes(user) {
         !user ||
         user.policy.organization !== undefined,
     },
+    {
```
nmcharlton, line 118: We have specific policies for stakeholder management now (or at least we will when https://github.com/Greenstand/treetracker-database/pull/92 is merged): `manage_stakeholders` and `list_stakeholders`

src/models/FilterStakeholder.js
```diff
@@ -0,0 +1,148 @@
+// import { stringToSearchRegExp } from '../utilities';
+
+/*
+ * A simple model for grower filter
+ */
+
+export const ALL_ORGANIZATIONS = 'ALL_ORGANIZATIONS';
+export const ORGANIZATION_NOT_SET = 'ORGANIZATION_NOT_SET';
+
+export default class Filter {
```
nmcharlton, line 10: Should be `FilterStakeholder`

src/models/FilterStakeholder.js
```diff
@@ -0,0 +1,148 @@
...
+  constructor(options) {
+    Object.assign(this, options);
+  }
+
+  getWhereObj() {
```
nmcharlton, line 15: Are we using LoopBack query syntax with the new API?

src/models/FilterStakeholder.js
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        if (this.lastName) {
+          where.lastName = lastName[0];
+        }
+        where.orgName = '';
+      } else {
+        // give all name fields the value for filter
+        where.orgName = this.name;
+        where.firstName = firstName || this.name;
```
nmcharlton, line 36: Setting the `firstName` and `lastName` in the filter for a non-person doesn't seem right to me.

nmcharlton: I wonder if the logic for constructing a name filter is the sort of more sophisticated task we should delegate to the API.

nmcharlton (commented):
Great work @gwynndp, this is looking great.

My main comment is to ensure we include the module name (_Stakeholder(s)_) in the names of components that are specific to that module. I understand omitting it from the filename for brevity, but it should be in the component name.

src/models/FilterStakeholder.js
gwynndp: I wasn't sure if we should keep it uniform from the client, so I did, but we don't have to.  There is validation for what can be filtered on the API as well.

src/components/Stakeholders/Table.js
gwynndp: It is the count of the main (logged in) stakeholder rows, not including their children, but you're right I'll revisit

src/context/AppContext.js
gwynndp: I added those policies with the current implementation but I think we'll have to revisit after the new policies are merged.

Issue #194, opened 2021-11-13
Resized webp images served in Admin panel: please add link to orignial file


camwebb:
Situation: the images in the Admin panel are of the form: `https://cdn.statically.io/img/treetracker-production-images.s3.eu-central-1.amazonaws.com/f=auto,w=640,h=535.5/2021.11.03.08.29....jpg` They are served as webp images. This is inconvenient when one is trying to link to and save the original JPEG image from Amazon S3. I discovered that if you remove the `/f=auto,w=640,h=535.5/` part, then the original image is presented. I can see that having a resized image is valuable for the fixed size admin panel. 

Request: somewhere on the admin panel popup, please add a (symbol with?) the _direct link to the full size jpeg_. Thanks.

nmcharlton:
That's a great suggestion. The resizing is mostly there to improve performance, but a link to the original image would be a sensible addition.

bedgarone:
Hi @nmcharlton, as a way to get started, I would like to solve this issue. Could you help me by describing the steps I should take?

After a quick walk through the code and the local app, I think this issue relates to the _Captures_ page. Being it mainly a table with all the Trees listed in text rows, I found that the _CaptureDetailDialog.js_ should be the target, since it makes the Capture image pop up after the capture is selected - in an _OptimizedImage_, which receives the original image (_renderImage.imageUrl_) as parameter. This issue demands that somewhere on the page we should include a icon/link to this referred url. 

Since there is no mockup, I'm not sure how to handle this change. 
- It could be added as a link or as a copy feature in the right panel: 
 <img width="223" alt="imagem" src="https://user-images.githubusercontent.com/33256390/142763645-cc2f38fd-8341-4acf-abfa-0f3caffc73b9.png">

- It could be over the image (with an absolute positioning) or next to the image. 

Please let me know if I should have a different approach, and, if possible, complement with what I might have missed. I woud also like to know how you're dealing with branching, shall I create a new one to later PR, and what should be the repository-related steps that I should take. Thank you, and very happy to enroll in this team!

nmcharlton:
Thanks @bedgarone. That's the right place to make the change and I like both of your proposals for a solution. The only additional thing I'd suggest is a clickable link to the original image, perhaps to open in a new tab.

As for the workflow, create a new branch on your fork for the feature and a draft PR that you can push to. That makes it easier to get feedback on ideas as you go along.

camwebb:
[...]

PR #206, opened 2021-11-23
feat: link to original image file

bedgarone:
## Description

The goal of this issue #194 is to provide a way for the user to access the _Capture_'s image in its original size, by opening it in a new tab or copying its url.

**What kind of change does this PR introduce?**
- Enhancement

## Issue
Each capture row pops up an optimized image (resized to improve performance) and there isn't a path to reach the original image in its full size.  

**What is the new behavior?**
The user is now able to either copy the original image's url or open the image in a new tab. 

**Steps taken:**
- Added a new item to the _Capture Data_'s array, which adds another item (the original image url) in the right panel;
- Set that itemps 'copy' property to true, so that a copy item appears and enables the user to copy the url;
- Added the same link to the original image over the optimized image, with _absolute_ positioning.

### Screenshots
imagem
imagem
imagem

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -7,17 +7,15 @@ import Grid from '@material-ui/core/Grid';
 import Divider from '@material-ui/core/Divider';
 import Chip from '@material-ui/core/Chip';
 import IconButton from '@material-ui/core/IconButton';
-import Snackbar from '@material-ui/core/Snackbar';
 import Drawer from '@material-ui/core/Drawer';
 import Box from '@material-ui/core/Box';
 import Close from '@material-ui/icons/Close';
-
-import FileCopy from '@material-ui/icons/FileCopy';
-import CloseIcon from '@material-ui/icons/Close';
 import OptimizedImage from './OptimizedImage';
 import LinkToWebmap from './common/LinkToWebmap';
 import { verificationStates } from '../common/variables';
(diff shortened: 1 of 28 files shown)


bedgarone:
@nmcharlton these were simple implementations of the discussed ideas. I'm would need some opinion on the link over the image, isn't it too distracting? Perhaps the link in the right panel is sufficient? 
Let me know if any of these need changes. Also, if the link over the image is going ahead, some refactor is needed in the style as I merely used strings while I suspect some items should use theme values (like the color, etc.).

nmcharlton:
@bedgarone This looks great! The link and copy option in the side panel as you've proposed work well.
We can remove the link from the image; I agree it's a bit distracting.

Nice work.

src/components/CaptureDetailDialog.js
```diff
@@ -313,6 +340,14 @@ function CaptureDetailDialog(props) {
           style={{ maxWidth: '100%' }}
           fixed
         />
+        <Link
```
nmcharlton, line 343: Let's remove this link from the image.

nmcharlton (changes requested):

src/components/CaptureDetailDialog.js
bedgarone: Hey @nmcharlton, just removed that link.

nmcharlton (approved):
Nice work

PR #207, opened 2021-11-24
feat: add copy button for Device Identifier in Grower Details

### LINKED ISSUE #145: Planter details is missing device identifier and the filter as per #116
Planter details panel on right side in the Planter section needs the Device identifier and then the link to the filtered device identifier view also see #116

tranquanghuy0801:
Resolves #145 

The search field `Device ID` has been implemented in the `master` branch to find the related growers based on `Device Identifier`

I only added the code to support copying the value of `Device Identifier` in Grower Details

### DIFF
--- a/src/components/GrowerDetail.js
+++ b/src/components/GrowerDetail.js
@@ -97,7 +97,13 @@ const GrowerDetail = (props) => {
               setGrowerRegistrations(sortedRegistrations);
               setDeviceIdentifiers(
                 sortedRegistrations
-                  .map((reg) => reg.device_identifier)
+                  .map((reg) => ({
+                    id: reg.device_identifier,
+                    os:
+                      reg.manufacturer.toLowerCase() === 'apple'
+                        ? 'iOS'
+                        : 'Android',
+                  }))
                   .filter((id) => id),
               );
             }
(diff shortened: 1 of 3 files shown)


src/components/GrowerDetail.js
```diff
@@ -19,6 +19,8 @@ import { GrowerContext } from '../context/GrowerContext';
 import EditGrower from './EditGrower';
 import OptimizedImage from './OptimizedImage';
 import LinkToWebmap from './common/LinkToWebmap';
+import { CopyButton } from './CaptureDetailDialog';
```
nmcharlton, line 22: `CopyButton` needs to be moved to the common components area rather than shared across components like this.

src/components/GrowerDetail.js
```diff
@@ -19,6 +19,8 @@ import { GrowerContext } from '../context/GrowerContext';
...
+import CopyNotification from './CopyNotification';
```
nmcharlton, line 23: Move this to `src/components/common/` as it's a component shared across other components.

nmcharlton (changes requested):

nmcharlton (approved):

PR #212, opened 2021-11-25
feat: add multi-select dropdown for verification status of captures f…

### LINKED ISSUE #114: Create multi-select dropdown for Verification Status so that Captures view total counts are correct
Total number of filtered not showing (only displaying total number of all trees) see screenshot
Screenshot 2021-06-11 at 10 48 43

It should be around 2000 captures in this case. 
We could also rename that entire section to captures.

tranquanghuy0801:
## Description

Add Multi-select dropdown for Verification Status field in filter of Captures page

**Issue(s) addressed**
- Resolves #114 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

The current design is a single dropdown menu for Verification Status field in Captures page

**What is the new behavior?**

The new design is a multi-select dropdown menu for Verification Status field in Captures page

## Breaking change

No breaking changes

## Video Screenshot 

https://user-images.githubusercontent.com/29462498/143463655-7004774c-67bd-4376-a601-d3e6d4994ca7.mov

### DIFF
--- a/src/components/CaptureFilter.js
+++ b/src/components/CaptureFilter.js
@@ -125,7 +125,7 @@ function Filter(props) {
   };
 
   const handleVerificationStatusChange = (event) => {
-    const value = event.target.value;
+    let value = event.target.value;
     let status = [];
     if (
       value[value.length - 1] === 'all' ||
@@ -136,6 +136,7 @@ function Filter(props) {
           ? []
           : verificationStatesArr
       );
+      value = verificationStatesArr;
     } else {
       setVerificationStatus(value);
     }


src/components/CaptureFilter.js
```diff
@@ -119,6 +132,59 @@ function Filter(props) {
...
+      setActive(undefined);
+      setApproved(undefined);
+      return;
+    }
+    handleChangeActiveApproved(value);
+    setVerificationStatus(value);
+  };
+
+  // set active and approved field based on the checkbox values
+  function handleChangeActiveApproved(value) {
```
nmcharlton, line 155: Unfortunately, this logic doesn't cover all possible combinations, and it's hard to follow what it's trying to do.
If _Approved_ and _Rejected_ are selected, everything (including _Awaiting Verification_) is included.

I'd suggest replacing `approved` and `active` in the filter model with these Verification States and only converting when the API query is constructed.
You will need to apply an `and` clause to the query object: https://loopback.io/doc/en/lb4/Where-filter.html#and--or

nmcharlton (changes requested):

nmcharlton:
Hey @tranquanghuy0801. What's the latest on this PR? I notice you've pushed a couple of commits, but the problem with the logic is still present. Do you need some help working out what needs to change?

tranquanghuy0801:
@nmcharlton The latest PR is just resolving some issues when pulling the updates from master branch. I understand how to fix and am working on this. Sorry for the delay.

tranquanghuy0801:
@nmcharlton I have updated the code as you suggested. Now you can check my work. Thanks

nmcharlton (changes requested):
@tranquanghuy0801 This seems to be working well, but I don't think it makes sense to be able to select none of the values (and get zero results).

Having nothing selected should return the same results as having all values selected (i.e., all captures).

tranquanghuy0801:
> @tranquanghuy0801 This seems to be working well, but I don't think it makes sense to be able to select none of the values (and get zero results).
> 
> Having nothing selected should return the same results as having all values selected (i.e., all captures).

Yeah I intended to show all results when the verification status is empty. It turned out it has an error in the function onHandleChange the value of these statues, which causes the incorrect data on the table. Now I fix that one.

nmcharlton (commented):
Thanks @tranquanghuy0801!

nmcharlton (approved):

PR #227, opened 2021-12-02
feat: add options for any-tag-set and tag-not-set

### LINKED ISSUE #199: Filter captures by 'Any tag'
In the Verify tool, users can currently filter by tag using either a specific tag name or _Not set_.
It should be possible to show all captures that have _any_ tag set.
This can be called _Any tag_ and sit below _Not set_ in the list.
(See `src/components/FilterTop.js`).

141854252-5424a4fd-dd0b-4026-a338-588f1ceaeb05

From a UX perspective, it would be nice (though not essential) to place a divider beneath the special values at the top of the list to make it clear where the actual list starts, or to apply distinct styling to the special values. Suggestions are welcome.

bstetzer32:
## Description

Added ability to filter by trees that have any tag set by passing in an id of "0" (see [treetracker-admin-api PR #608](https://github.com/Greenstand/treetracker-admin-api/pull/608)). 
**Issue(s) addressed**

- Resolves #199

- [x] Enhancement

## Issue

Users only have the ability to filter by either tag-not-set or by an individual tag.

**What is the new behavior?**
Users now have the ability to filter by any-tag-set as well as tag-not-set or by an individual tag. 

## Breaking change

No

## Other useful information
I included 2 possible options for a divider: one using the renderOption prop on AutoComplete to apply a class to the option if it is any-tag-set, and the other using a CSS style for #tag-option-1.


src/components/FilterTop.js
```diff
@@ -375,18 +385,20 @@ function Filter(props) {
...
-                  }
+                  setTag(newVal);
                 }}
                 onInputChange={(_oldVal, newVal) => {
                   setTagSearchString(newVal);
                 }}
                 renderInput={(params) => {
                   return <TextField {...params} label="Tag" />;
                 }}
+                // Option B for divider
```
nmcharlton, line 396: If Material UI doesn't provide a good way of adding dividers, let's go without for now.

nmcharlton (changes requested):
Hi @bstetzer32, sorry for the delay in reviewing this – I've got quite a backlog to work through and time is limited.

I've noticed that the API is returning duplicate captures.
If you filter Approved captures with Any tag set, you'll see capture 951788 twice.
This is also reflected in the capture counts which don't quite add up.
I'll raise an issue against the API that we'll need to fix.

src/components/FilterTop.js
```diff
@@ -68,6 +69,9 @@ const styles = (theme) => {
     noSpecies: {
       fontStyle: 'italic',
     },
+    // tagDecorator: {
```
pieroapretto, line 72: Is this comment needed?

nmcharlton (approved):
Looking good

PR #228, opened 2021-12-03
fix: display Full Capture Details in Verify and Captures pages

### LINKED ISSUE #220: Full capture details not loaded in Capture Detail dialog
When a user opens the Capture Detail dialog (by clicking the tree icon in the _Verify_ tool or a row in the _Captures_ tool), the Admin Panel should load the full capture details, including tags and species.

Screenshot 2021-11-29 at 19 36 16

**Captures**
In the _Captures_ tool, Species and Additional tags are missing.
The client appears to query the capture but not these additional details.
Screenshot 2021-11-29 at 19 48 50

**Verify**
In the _Verify_ tool, all of the "Other" tags are missing, as well as Species and token status.
The client does not query the capture at all.
Screenshot 2021-11-29 at 19 41 44

I can see a couple of things causing this:

- We're not wrapping these views with `CaptureDetailProvider` tags
- We don't return `capture` from `CaptureDetailContext::getCapture()` so it can't be passed to subsequent queries

tranquanghuy0801:
…Captures pages

## Description

This PR fixed the issue of not displaying full species and other tags in Capture Detail Dialog.

**Issue(s) addressed**

- Resolves #220 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Not display the species tags in Captures page, and not display the species tags and the other tags in Verify page

**What is the new behavior?**

Both species and other tags are displayed fully in Verify and Captures pages.

## Breaking change

No breaking change

### DIFF
--- a/src/api/treeTrackerApi.js
+++ b/src/api/treeTrackerApi.js
@@ -1,7 +1,7 @@
 import { handleResponse, handleError, getOrganization } from './apiUtils';
 import { session } from '../models/auth';
 
-const FIELDS = {
+const CAPTURE_FIELDS = {
   uuid: true,
   imageUrl: true,
   lat: true,
@@ -21,7 +21,7 @@ const FIELDS = {
   captureApprovalTag: true,
   rejectionReason: true,
   note: true,
-}
+};
 
 export default {
   getCaptureImages(
@@ -42,7 +42,7 @@ export default {
       order: [`${orderBy} ${order}`],
       limit: rowsPerPage,
       skip,
-      fields: FIELDS,
+      fields: CAPTURE_FIELDS,
     };
 
     const query = `${
@@ -122,15 +122,9 @@ export default {
       .catch(handleError);
   },
   getCaptureById(id) {
-    const lbFilter = {
-      order: [`${'asc'} ${'id'}`],
-      fields: FIELDS,
-    };
-
-    const paramString = `filter=${JSON.stringify(lbFilter)}`;
     const query = `${
       process.env.REACT_APP_API_ROOT
-    }/api/${getOrganization()}trees/${id}${paramString ? '?' + paramString : ''}`;
+    }/api/${getOrganization()}trees/${id}`;
     return fetch(query, {
       headers: {
         Authorization: session.token,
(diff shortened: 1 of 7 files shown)


src/components/CaptureDetailDialog.js
```diff
@@ -85,7 +88,10 @@ function CaptureDetailDialog(props) {
 
   useEffect(() => {
     cdContext.getCaptureDetail(capture.id);
-
+    const filterSpecies = capture.speciesId ? speciesContext.speciesList.filter(spec => spec.id === capture.speciesId) : [];
```
nmcharlton, line 91: Instead of `speciesList.filter()` here, use `speciesList.find()` with the same comparison function.
It will return the first match, or `null` if there are no matches.

src/api/treeTrackerApi.js
```diff
@@ -113,9 +122,15 @@ export default {
       .catch(handleError);
   },
   getCaptureById(id) {
+    const lbFilter = {
+      order: [`${'asc'} ${'id'}`],
```
nmcharlton, line 126: `order` doesn't make sense here. This query only returns one record.

nmcharlton (changes requested):
This is good, but it doesn't address the whole problem.
In particular, we still don't see the capture tags (AKA treeTags).
See my comments in the issue for how to resolve that part:
> I can see a couple of things causing this:
>
> - We're not wrapping these views with `CaptureDetailProvider` tags
> - We don't return capture from `CaptureDetailContext::getCapture()` so it can't be passed to subsequent queries

src/api/treeTrackerApi.js
```diff
@@ -1,6 +1,28 @@
 import { handleResponse, handleError, getOrganization } from './apiUtils';
 import { session } from '../models/auth';
 
+const FIELDS = {
```
nmcharlton, line 4: I'd suggest renaming this `CAPTURE_FIELDS` to make its use clear

src/api/treeTrackerApi.js
nmcharlton: In fact, this endpoint ignores the param string, so the filter is redundant here. The whole record is always returned for a query by ID.

nmcharlton (approved):
@tranquanghuy0801 I hope you don't mind, I've made the changes myself so we can get the fix into the next build. Thanks for your work on this!

tranquanghuy0801:
> @tranquanghuy0801 I hope you don't mind, I've made the changes myself so we can get the fix into the next build. Thanks for your work on this!

@nmcharlton That's alright. I totally understand. Happy to have a new release.

PR #229, opened 2021-12-03
fix: tag search in Verify and Captures filters

### LINKED ISSUE #221: Verify and Captures filters: Tag search not working
When a user types a search string into the _Tags_ filter box in both the _Verify_ and _Captures_ tools, the client should query tags containing the search value from the API:

Screenshot 2021-11-29 at 20 45 46

The query is only currently triggered when the search string is empty (in both `FilterTop.js` and `CapturesFilter.js`:
```
  useEffect(() => {
    if (tagSearchString === '') {
      const abortController = new AbortController();
      tagsContext.getTags(tagSearchString, { signal: abortController.signal });
      return () => abortController.abort();
    }
  }, [tagSearchString]);
```

Admittedly, the tags filter solution is a little overengineered, so we could remove the real-time API querying and just query on the client, but the simple solution is to re-enable the real-time queries.

tranquanghuy0801:
## Description

Verify and Captures Filters: tag search not working

**Issue(s) addressed**

- Resolves #221 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Tag search is only executed when the input tag changes and the tag is an empty string

**What is the new behavior?**

Tag search will be executed when the input tag changes

## Breaking change

No

### DIFF
--- a/src/components/CaptureFilter.js
+++ b/src/components/CaptureFilter.js
@@ -102,11 +102,9 @@ function Filter(props) {
   const [tokenId, setTokenId] = useState(filter?.tokenId || filterOptionAll);
 
   useEffect(() => {
-    if (tagSearchString === '') {
-      const abortController = new AbortController();
-      tagsContext.getTags(tagSearchString, { signal: abortController.signal });
-      return () => abortController.abort();
-    }
+		const abortController = new AbortController();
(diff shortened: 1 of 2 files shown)


nmcharlton (approved):
Thanks @tranquanghuy0801!

PR #237, opened 2021-12-08
feat: add capture tags to Captures table

### LINKED ISSUE #193: Add capture tags to Captures list
We need to be able to see species, tags and other information in the list.

tranquanghuy0801:
## Description

Add Capture Tags to Captures table

**Issue(s) addressed**

- Resolves #193 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

Don't have the Capture Tags column in Captures table

**What is the new behavior?**

New Capture Tags column in Captures table

## Breaking change

No

### DIFF
--- a/src/components/Captures/CaptureTable.js
+++ b/src/components/Captures/CaptureTable.js
@@ -103,21 +103,24 @@ const CaptureTable = () => {
 
   useEffect(() => {
     populateSpeciesLookup();
+  }, [speciesContext.speciesList]);
+
+  useEffect(() => {
     populateTagLookup();
-  }, [filter]);
+  }, [tagsContext.tagList]);
 
   useEffect(async () => {
-    console.log(`Captures (${captures.length}): ${JSON.stringify(captures)}`);
-
     // Don't do anything if there are no captures
     if (!captures?.length) {
       return;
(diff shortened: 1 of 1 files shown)


Siddhant-K-code (approved):
LGTM 🚀

nmcharlton:
@tranquanghuy0801 This looks good. The only thing missing is _Additional tags_, which aren't part of the `trees` LoopBack model, so we have to request them separately using `treeTrackerApi::getCaptureTags()`.
It would be good to try to add these for the listed captures and see what the performance is like.

If that doesn't work well, we may need to modify the API to always return those tags with the capture.

nmcharlton:
> @tranquanghuy0801 This looks good. The only thing missing is _Additional tags_, which aren't part of the `trees` LoopBack model, so we have to request them separately using `treeTrackerApi::getCaptureTags()`. It would be good to try to add these for the listed captures and see what the performance is like.
> 
> If that doesn't work well, we may need to modify the API to always return those tags with the capture.

@tranquanghuy0801 I can take a look at adding in the Additional Tags to the list.

nmcharlton:
image

PR #249, opened 2021-12-21
Fix: previous capture details

### LINKED ISSUE #241: Captures tool: Previous capture details are displayed briefly in Capture Details dialog
## Tool
Captures

## Steps to reproduce

1. Click any record to bring up the _Capture Details_ dialog. After a few moments, all of the data for the capture is loaded.
2. Close the dialog and select another capture. The previously selected capture's details are displayed briefly until the new one loads.

https://user-images.githubusercontent.com/5558838/145723251-9edda824-ad61-4d85-bb64-19f2740bd5f9.mov

## Desired behaviour

The previously selected capture does not appear.
See how this is done in the Verify tool for comparison - we set the value of `capture` passed to the dialog to `null` when it is closed.
Ideally, we'd wait for the animation to complete before clearing the capture, but this isn't too important.

## Relevant files/components

src/components/CaptureDetailDialog.js
src/components/Captures/CaptureTable.js

bedgarone:
## Description

To avoid the previous capture details being displayed briefly in Capture Details dialog, the capture variable is set to an empty object when the Dialog is closed (on the closeDrawer handler method).

**Issue(s) addressed**

- Resolves #241 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

When opening a capture dialog, the details of the previous seen capture (if applicable) are displayed briefly before being updated.

**What is the new behavior?**
When opening a capture dialog, the previous seen capture won't appear since it was cleared from the capture variable when closing the previous capture dialog. The new capture is shown since the beginning of the dialog opening, after loading. 

## Breaking change

No

### DIFF
--- a/src/components/Captures/CaptureTable.js
+++ b/src/components/Captures/CaptureTable.js
@@ -101,10 +101,6 @@ const CaptureTable = () => {
   const [captureTagLookup, setCaptureTagLookup] = useState({});
   const [isOpenExport, setOpenExport] = useState(false);
   const classes = useStyle();
-  const [captureDetail, setCaptureDetail] = useState({
-    isOpen: false,
-    capture: {},
-  });
 
   useEffect(() => {
     populateSpeciesLookup();
@@ -155,10 +151,6 @@ const CaptureTable = () => {
   const toggleDrawer = (id) => {
     getCaptureAsync(id);
     setIsDetailsPaneOpen(!isDetailsPaneOpen);
-    setCaptureDetail({
-      isOpen: true,
-      capture: capture,
-    });
   };
 
   const createToggleDrawerHandler = (id) => {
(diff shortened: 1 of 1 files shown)


bedgarone:
@nmcharlton As suggested by you, the capture value should be set to an empty object when the Dialog is closed, thus, on the _closeDrawer_ handler method. However, the capture variable is set on the CapturesContext and the set method is there. Apparently, exporting that setCapture method and use it in handler does the job. Since I'm not sure if I should edit the CapturesContext.js file, please let me know if a different approach should be made.

Also, I would like to try setting it to null only after the animation is complete, as you suggested. However, the animation is so quick that I doubt I'll be able to verify and notice any change when trying to accomplish that.

nmcharlton:
@bedgarone I think the fix you've proposed is perfectly fine, and it will enable a fix for #241.

nmcharlton (approved):
Nice work

Issue #251, opened 2021-12-22
Commas are missing to describe thousands place


ZavenArra:
Please always insert a comma to show the thousands, millions, and billions place in all numbers as a standard.  This card shows one example of a missing command.  A number like 1200 should also be rendered as 1,200.

![image](https://user-images.githubusercontent.com/364622/147166591-884d6b03-750e-492d-ace9-5d6447be9566.png)

arcadioramos:
Hi @dadiorchen , I would like to give it a try to this but I can't login to check the app I see that the credentials are in a slack channel but idk how to get into that slack channel

dadiorchen:
I will add you to that channle

dadiorchen:
I have added you into the slack, also I sent you a email with the dev env credential

nmcharlton:
@arcadioramos Thanks for picking this up. Make sure to use `countToLocaleString()` in src/common/numbers.js to ensure the correct separator is used in different locales.

arcadioramos:
Thank you @dadiorchen @nmcharlton

nmcharlton:
Fixed in #255

PR #255, opened 2021-12-25
fix: numbers were not  using commas on dashboard screen due a bug

arcadioramos:
## Description

I've implemented the function countToLocaleString(number) to show big numbers separated with commas, also I found another small bug on the dashboard and it was that this number:
image
was a String so it was not applying the function correctly due to the type of the variable used so I parsed that variable to work as it is supposed to work.

**Issue(s) addressed**

- Resolves the issue  Commas are missing to describe thousands place #251

- [x] Bug fix

## Issue

image
Commas were missing on big numbers.

**What is the new behavior?**
Actually, the dashboard represents the numbers with commas on millions and billions place in all numbers as a standard.
image

## Breaking change


src/common/numbers.js
```diff
@@ -1,4 +1,6 @@
-export function countToLocaleString(num) {
+const countToLocaleString = (num) => {
   if (!num?.toLocaleString || typeof num !== 'number') return num;
   return num.toLocaleString();
-}
+};
+
+export { countToLocaleString };
```
dadiorchen, line 6: auto-formating?

arcadioramos: Yes, I think that is what happened there and I also changed that function to an arrow function

nmcharlton (approved):
Nice work!

Issue #250, opened 2021-12-22
Round up to whole numbers on all statistic cards


ZavenArra:
This statistic card shows tenths, hundredths, and thousands place after the decimal.  Instead round up to a whole number.   This can be fixed on the API side as well.

![image](https://user-images.githubusercontent.com/364622/147166487-6d852f8a-a0f5-45a6-ad63-fb650b8dff16.png)

sunnyjain15699:
Hi! I would like to work on this issue. Can you guide me through it ?

dadiorchen:
Yes, please read the `readme` to get started, installing and running the project locally, then try to reproduce the problem, and fix it. 

If you want to join our slack community, please give me you email address, then I can invite you

dadiorchen:
@arcadioramos  is this the same case as your last PR? using countToLocaleString? If it's true, I think it would be better you make a quick PR on this so we can solve this issue soon.

dadiorchen:
https://github.com/Greenstand/treetracker-admin-client/issues/251

arcadioramos:
Hi @dadiorchen , it is a different case because here we have that the numbers have decimal and here what I can see is that they want to remove the decimal and to add +1 if tur number has a decimal to round the number up, I think I can solve it quick but give me until tomorrow because I'm kinda busy at the moment

PR #257, opened 2021-12-29
Fix: rounded numbers in cards

arcadioramos:
## Description

I had to add the javascript function Math.ceil() on the component ReportingCard.js to round up to whole numbers on all the cards and I also did this for the function allocated in the numbers.js file

**Issue(s) addressed**

- Resolves This issue solves Round up to whole numbers on all statistic cards #250

- [x] Bug fix

## Issue

Numbers were displaying with decimals and this wasn't the expected behaviour.

**What is the new behavior?**

## Breaking change
Numbers are rounded up to only display whole numbers :) 

## Other useful information

Screenshots: 
Before: 
image
After: 
image


dadiorchen:
@arcadioramos because some reason, I need to copy your code and push to another PR, so I now close your this PR

PR #261, opened 2021-12-31
Keycloak auth

### LINKED ISSUE #1027: Keycloak: take example and integrate it into Admin Panel (need cloud, api, react support)

TANguyen1893:
## Description

Integrate the admin panel with keycloak. Use keycloak as an identtiy provider instead of implementing our own auth services.

https://docs.google.com/document/d/1HtZm_ccs-tZwcmjobj1Ks4uQlovA7_47v5hIM6HDLGY 

**Issue(s) addressed**

- Resolves #1027 

- [x] Enhancement

## Issue

We implement our own login system, session management, and user management.

**What is the new behavior?**
Keycloak will provide security features out of the box so we don't need to re-implement them.

## Breaking change

Yes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "treetracker-admin-client",
-  "version": "1.121.1",
+  "version": "1.124.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "treetracker-admin-client",
-      "version": "1.121.1",
+      "version": "1.124.0",
       "dependencies": {
         "@date-io/date-fns": "^1.3.13",
         "@material-ui/core": "^4.9.10",
@@ -36,7 +36,7 @@
         "i": "*",
         "import": "*",
         "install": "*",
-        "keycloak-js": "^16.1.0",
+        "keycloak-js": "^21.0.1",
         "loglevel": "^1.6.3",
         "material-ui-chip-input": "^2.0.0-beta.2",
         "mdi-material-ui": "*",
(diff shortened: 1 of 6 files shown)


dadiorchen:
@TANguyen1893 do we need to do some change on the API server side?

src/components/DashStat.container.js
```diff
@@ -12,17 +12,20 @@ import PeopleOutlineOutlinedIcon from '@material-ui/icons/PeopleOutlineOutlined'
...
 
   const [total, setTotal] = useState(null);
-
+  const {
+    keycloak: { token },
+  } = useKeycloak();
   const getTotal = async () => {
     console.log('-- dash getTotal');
-    const { count } = await api.getCaptureCount(activeFilter);
+    const { count } = await api.getCaptureCount(activeFilter, token);
```
nmcharlton, line 28: I don't think the component should need to know about `token` and be responsible for retrieving it.
This may be better placed in the API function.

dadiorchen: @TANguyen1893 is using a hook for this, so it's easy to get from the component side.

dadiorchen: To make this structure better, maybe we can use custom hook to initial API object

src/models/auth.js
```diff
@@ -15,36 +11,34 @@ const POLICIES = {
...
-    return user.policy.policies.some((r) => r.name === p) ? true : false;
-  }
+function hasPermission(permissions) {
+  const {
+    keycloak: { tokenParsed },
+  } = useKeycloak();
+  const roles = tokenParsed?.realm_access?.roles;
+  return permissions
+    .map((permission) => roles?.includes(permission))
+    .reduce((a, b) => a && b);
```
dadiorchen, line 21: @TANguyen1893 I think it might be limited to just using role to check the permission? Seems we need a permission-role-policy pattern to define a fine-grained system? like:
policy A defined that role B has the permission to doing action C
like: 
The tree-user has permission to view the list of tree
The tree-manager has the permission to edit the list of tree

TANguyen1893: @dadiorchen You are right. It would be much better to have a fine-grained permission system. I would prefer that than just simply roles.

PR #273, opened 2022-01-08
Improve capture multi-selection UX in for batch verification

### LINKED ISSUE #233: Improve capture multi-selection UX in for batch verification
Currently, when a user wishes to select multiple captures in the Verify tool to perform a batch approval/rejection/tagging (a very common use case), they must either:

- Select the first in a range, hold shift and select the last (range selection); or
- Hold the ctrl/cmd key and select each capture individually to add to the selection

The second of these in particular is a clunky experience for the user. When holding ctrl on Windows and trying to scroll, the browser can zoom in and out instead. Therefore, the user has to release the ctrl key to scroll, making it more likely they'll accidentally select another capture without the ctrl key held and inadvertently clear their original selection.

There is also no feedback on the number of captures selected for batch processing.

image

The following improvements should be made to the UX:

- Display the number of captures selected
- Always display the selection checkbox for captures to signal that clicking the capture will check/uncheck the box
- Clicking a capture should toggle its selection, rather than replacing the existing selection

Here's an initial design for the UI:
Admin Panel - Bulk Select Captures - Greenstand-v01 01-1

I'd suggest that the _Selected Captures_ dialog go above the _APPROVE | REJECT_ switch as it applies to both options.

tranquanghuy0801:
## Description

Improve capture multi-selection UX in for batch verification

**Issue(s) addressed**

- Resolves #233 

- [X] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

Users have to press shift/Cmd to be able to select multiple captures to verify. Additionally, clicking the selected capture led to another capture getting selected, which is not appropriate. Side panel on the right did not display the total number of selected captures on the current page.

**What is the new behavior?**

Users can select multiple captures without using shift/Cmd, any capture can be toggled to be selected/un-selected. Additionally, the side panel on the right displays the total number of selected captures on the current page, users also can have the option to `Select All / Unselect All` captures on the right side panel.

## Breaking change

No

## Other useful information

Screenshot 2022-01-09 at 21 58 21

### DIFF
--- a/src/context/VerifyContext.js
+++ b/src/context/VerifyContext.js
@@ -197,6 +197,7 @@ export function VerifyProvider(props) {
         });
       setCaptureImagesSelected(captureShiftSelected);
     } else {
+      setCaptureImageAnchor(captureId);
       setCaptureImagesSelected({
         ...captureImagesSelected,
         [captureId]: !captureImagesSelected[captureId],


src/context/VerifyContext.js
```diff
@@ -247,53 +259,53 @@ export function VerifyProvider(props) {
...
-    }
-    //}}}
-  };
+  // const clickCapture = (payload) => {
+  //   //{{{
+  //   const { captureId, isShift, isCmd, isCtrl } = payload;
+  //   if (!isShift && !isCmd && !isCtrl) {
+  //     setCaptureImagesSelected([captureId]);
+  //     setCaptureImageAnchor(captureId);
+  //   } else if (isShift) {
```
nmcharlton, line 268: Can we try to keep the shift-click functionality working?

src/components/SidePanel.js
```diff
@@ -106,6 +132,32 @@ function SidePanel(props) {
...
+          >
+            Select All
+          </Button>
+          <Button
+            color="primary"
+            variant="contained"
+            className={classes.selectButton}
+            onClick={() => setSelectedCaptures(false)}
+          >
+            Unselect All
```
nmcharlton, line 158: I think "Select None" might be better

nmcharlton (changes requested):

tranquanghuy0801:
@nmcharlton  I have updated the code as your suggestions. The only thing I did not change is that I still use the `parseInt()` in the function `getCaptureSelectedArr` in `VerifyContext` as the `Object.keys()` only accepts keys as strings, thus returning array of strings and we need array of integers for processing the `approveAll`.

src/context/VerifyContext.js
```diff
@@ -154,19 +156,67 @@ export function VerifyProvider(props) {
...
+        .slice(
+          Math.min(indexAnchor, indexCurrent),
+          Math.max(indexAnchor, indexCurrent) + 1
+        )
+        .forEach((capture) => {
+          captureShiftSelected[capture.id] = true;
+        });
+      setCaptureImagesSelected(captureShiftSelected);
+    } else {
+      setCaptureImagesSelected({
```
nmcharlton, line 200: You need to call `setCaptureImageAnchor(captureId);` here, otherwise shift-click doesn't work properly.

src/components/SidePanel.js
```diff
@@ -70,6 +84,18 @@ function SidePanel(props) {
     setRejectionReason(DEFAULT_REJECTION_REASON);
   }
 
+  function setSelectedCaptures(value) {
+    if (value) {
+      let captureSelected = {};
```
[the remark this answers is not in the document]
nmcharlton: @tranquanghuy0801 This one need to be addressed to.

nmcharlton (changes requested):

nmcharlton (approved):

PR #283, opened 2022-01-11
Add Log In cypress integration tests

PtrHtn:
## Description
Add integration tests for Login page

## Breaking change
No

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,10 @@ To run tests:
 
 `npm test`
 
+### Cypress
+
+Make your own `/cypress/fixtures/login.json` file containing the actual credentials in order to run the cypress tests.
+
 ## How to log
 
 We use [loglevel](<(https://github.com/pimterry/loglevel)>) for logging, with some conventions. Using loglevel, we will be able to open/close a single file's log by chaining the level of log on the fly, even in production env.


cypress/fixtures/login.json
```diff
@@ -1,3 +1,5 @@
 {
+  "user_name": "admin",
+  "password": "X4YRyeg7ES",
```
nmcharlton, line 3: Please don't include credentials in these scripts, unless they are dummy credentials from a mocked API or DB.

nmcharlton (changes requested):

cypress/fixtures/login.json
PtrHtn: I modified the script's credentials.

cypress/integration/login.spec.py.js
```diff
@@ -9,6 +9,31 @@ describe.only('Login', () => {
     loginPage.login('incorrectName', 'incorrectPassword');
     loginPage.expect_Displayed(login.error_message);
   });
+
+  it('logs in when "correct" credentials are entered', () => {
```
gwynndp, line 13: This test is failing for me.  We may need to mock the request response so it is successful without actual login info

gwynndp: Otherwise, looking good! :)

PtrHtn: Have you entered the proper user name and password into cypress/fixtures/login.json ?
  "user_name": "replace_with_the_correct_user_name",
  "password": "replace_with_the_correct_password",

gwynndp: I think we may want to offer a mocked login in the future.
  
For now can you please rename the login.json to login-example.json and update the .gitignore so that it won't ever commit the login.json file.  Then add to the README instructions for developers to create their own login.json with the actual creds in order to run the cypress tests.  Thanks!

gwynndp (approved):

nmcharlton (approved):

PR #286, opened 2022-01-13
fix: Fixed Menu Width

### LINKED ISSUE #198: The menu UI got a bit displaced because of the width
Screen Shot 2021-11-15 at 2 06 07 PM

Tiaraha-funny:
## Description

This branch is created to fix the width of the menu to have a fixed width on User manager, Account and Species components. 

**Issue(s) addressed**

- Resolves #198

- The menu wraps when resizing the screen on Account, Species and User Manager pages.
- Set the width of the menu and those pages to be fixed and enlarged it.
- Div element is used as a wrapper container instead of using Grid element. 

## Issue

**Previous behavior**
image

**What is the new behavior?**
menu

### DIFF
--- a/src/components/Home/Home.js
+++ b/src/components/Home/Home.js
@@ -100,7 +100,7 @@ function Home(props) {
           <Menu variant="plain" />
         </Paper>
       </Grid>
-      <Grid className={classes.rightBox} style={{ flexGrow: 1 }}>
+      <Grid className={classes.rightBox}>
         <Box className={classes.box2}>
           <Grid container spacing={5} className={classes.version}>
             <Grid item xs={3}>
--- a/src/components/Home/Home.styles.js
+++ b/src/components/Home/Home.styles.js
@@ -45,7 +45,7 @@ const styles = (theme) => ({
     right: 0,
     backgroundColor: 'rgb(239, 239, 239)',
     boxSizing: 'border-box',
-    overflowY: 'scroll',
+    overflowY: 'auto',
   },
   welcomeBox: {
     height: '100%',


src/components/Account.js
```diff
@@ -187,88 +187,84 @@ function Account(props) {
...
-          <Grid container className={classes.rightBox}>
-            <Grid item xs={12}>
-              <Grid container className={classes.titleBox}>
-                <Grid item>
-                  <AccountIcon className={classes.accountIcon} />
-                </Grid>
-                <Grid item>
-                  <Typography variant="h2">Account</Typography>
-                </Grid>
+    <div style={{ display: 'flex', flexDirection: 'row', flexWrap: 'nowrap' }}>
```
nmcharlton, line 190: Set this container to `width: '100%'` to ensure it fills the screen.
It's also better to create a class for these CSS properties – see line 23.

src/components/Account.js
```diff
@@ -187,88 +187,84 @@ function Account(props) {
...
+      <Paper elevation={3} style={{ width: '300px' }}>
```
nmcharlton, line 191: Since we want a fixed and consistent menu width throughout the tool, it's best to set the width of the outer element of the `<Menu>` component in Menu.js.
There's already a `MENU_WIDTH` value there, but it's only being applied to drawer version of the menu.
If we move the width to the core menu component, it will apply everywhere.

src/components/Account.js
```diff
@@ -187,88 +187,84 @@ function Account(props) {
...
+        <Menu variant="plain" />
+      </Paper>
+
+      <Grid item xs={9} style={{ width: '1000px' }}>
```
nmcharlton, line 195: Don't use a fixed width here.
Instead, set `flexGrow: 1` and remove the `xs={9}` property.
That, combined with the 100% width of the parent element will ensure the content fills the container.

src/components/SpeciesTable.js
```diff
@@ -235,86 +242,92 @@ const SpeciesTable = (props) => {
...
-                  </Grid>
+    <div
+      style={{
+        display: 'flex',
+        flexDirection: 'row',
+        flexWrap: 'nowrap',
+        height: '100vh',
+      }}
+    >
+      <Paper elevation={3} style={{ width: '300px' }}>
```
nmcharlton, line 253: See the comments against Account.js. These all apply here too.

src/components/Users.js
```diff
@@ -563,78 +563,73 @@ function Users(props) {
...
-        </Grid>
-        <Grid item xs={9}>
-          <Grid container className={classes.rightBox}>
-            <Grid item xs={12}>
-              <Grid
-                container
-                justify="space-between"
-                className={classes.titleBox}
-              >
+    <div
```
nmcharlton, line 566: See comments against Account.js - the same comments apply here.

nmcharlton (changes requested):
It would be best to set the fixed width on the Menu component itself, rather than everywhere it's used. This will keep its appearance consistent across the Admin Panel.

Let me know if you need any help with this.

nmcharlton:
This is coming along nicely. A few more suggestions:
1. Reduce the `marginRight` of `menuItem` – it's making the menu much wider than it needs to be
1. The menu panel on the Home page no longer extends to the bottom of the page
1. Use `flexGrow: 1` on the User Manager and Species page content to ensure the tables fill the width of the screen
1. Use `padding` around the content on User Manager and Species rather than a `gap` in the flex grid to ensure even spacing all the way round

I hope this makes sense.

nmcharlton (approved):
@Tiaraha-funny and @mitia-anah Thank you for all your work on this. I hope you don't mind, I applied a few final tweaks to your PR, but nothing major.

PR #304, opened 2022-01-19
chore: fix "account" integration test

PtrHtn:
## Description
_account page:_
_returns to the Login page after clicking the "LOG OUT" button_

## Breaking change
No breaking change


gwynndp (approved):

gwynndp:
I'm having difficulty seeing these tests pass.  Is there something I need to do in particular or could it be because something necessary is in a different PR?

Issue #297, opened 2022-01-16
Captures: show loading indicator


nmcharlton:
It can take several seconds for any records to appear in the Captures table, especially when a complex filter is applied.
We need to give the user feedback that loading has not completed, e.g., a spinner.

mitia-anah:
Hi @nmcharlton,

Would you mind providing us with some clarification on this task, please? For us to know how it should be. 
As we can see in the User Manager table, there is a loading indicator before the table appears. So, is the Captures table also expected to have the same features?

nmcharlton:
@mitia-anah That's right.

mitia-anah:
@nmcharlton Thank you. We are going to work on this issue.

PR #305, opened 2022-01-19
feat: Loading Indicator On Captures Page

mitia-anah:
## Description

- Added loading indicator to captures page. While the captures table is loading, the loading indicator is displayed. 

**Issue(s) addressed**

- Resolves #https://github.com/Greenstand/treetracker-admin-client/issues/297

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

- It takes several seconds for any records to appear in the Captures table,

**What is the new behavior?**
- The loading indicator is displayed to give the user feedback that loading has not been completed yet.

## Breaking change


TANguyen1893 (commented):
Would it be easy to add tests to ensure the changes that are made will not break anything?

nmcharlton:
@mitia-anah Please can you rebase or merge the latest Greenstand master branch into your feature branch?

src/components/Captures/CaptureTable.js
```diff
@@ -258,26 +259,35 @@ const CaptureTable = () => {
...
-                      speciesLookup,
-                      captureTagLookup[capture.id] || [],
-                      attr,
-                      renderer
-                    )}
-                  </TableCell>
-                ))}
-              </TableRow>
-            ))}
+          {captures.length === 0 ? (
```
nmcharlton, line 262: Unfortunately, this isn't quite enough. There may be occasions when zero captures match the filters, so we want to show an empty list.
Instead, we will need to add an `isLoading` boolean flag to CapturesContext that tracks whether a Captures load is in progress.

nmcharlton (changes requested):
I've merged in changes from another PR that got lost, so you'll need to do `git pull origin` again before making changes.

src/components/Captures/CaptureTable.js
```diff
@@ -120,7 +121,7 @@ const CaptureTable = () => {
   useEffect(async () => {
     // Don't do anything if there are no captures
     if (!captures?.length) {
-      return;
+      return isLoading;
```
nmcharlton, line 124: Nothing uses the return value from this function, so this doesn't do anything.
The `isLoading` flag should be set to `true` before the API is called and `false` when the response is received. See  VerifyContext for guidance:
https://github.com/Greenstand/treetracker-admin-client/blob/4498a21d77b7c02ab1647248df303a603a5da198/src/context/VerifyContext.js#L142

This flag should then be used in CapturesTable to determine whether to show the loading indicator.

nmcharlton (changes requested):

src/components/Captures/CaptureTable.js
```diff
@@ -258,26 +260,35 @@ const CaptureTable = () => {
...
-                      attr,
-                      renderer
-                    )}
-                  </TableCell>
-                ))}
-              </TableRow>
-            ))}
+          {captures.length === 0 ? (
+            <div
+              style={{ position: 'relative', left: '300%', bottom: '-179px' }}
```
nmcharlton, line 265: Move the style details to CaptureTable.styles.js as a new class that is referenced with an attribute like `className={classes.loadingIndicator}`.

nmcharlton, line 265: This positioning seems odd. Take a look at how `CircularProgress` is positioned in other admin panel tools.

src/components/Captures/CaptureTable.js
nmcharlton: This should be `{isLoading ? (`

nmcharlton (approved):
I've moved things around a little – it's a good idea for the Context to manage the loading state and the Table to just read it.

PR #321, opened 2022-01-25
fix issue of verify capture page loading when switching

### LINKED ISSUE #310: Verify: Changing page before the previous load has completed breaks capture loading
If you change pages two or more times in quick succession in the Verify tool, before the previous query has completed, captures do not load and loading stops working. At that point, you have to refresh the page to start again.

When the second page skip is selected, the `abortController` is activated on the first `api.getCaptureImages()` request in VerifyContext.js, causing the result to be undefined.
Unfortunately, we do not check the value is valid before attempting to log the length of the resulting array.
This causes an exception, and so the `isLoading` value is never set to `false`. Therefore, all subsequent queries are ignored, as loading is already in progress.

Now that we have a abort controller, we can do away with the logic that cancels the load if `isLoading` is `true`.

https://user-images.githubusercontent.com/5558838/150420978-092dc197-3799-4f8c-be67-4fed747e630e.mov

tranquanghuy0801:
## Description

fix issue of verify capture page loading when switching

**Issue(s) addressed**

- Resolves #310 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

If you change pages two or more times in quick succession in the Verify tool, before the previous query has completed, captures do not load and loading stops working

**What is the new behavior?**

change the loading query to only load the capture images when `isLoading` is set to `false`

## Breaking change

No


src/context/VerifyContext.js
```diff
@@ -58,8 +58,10 @@ export function VerifyProvider(props) {
   /* load captures when the page or page size changes */
   useEffect(() => {
     const abortController = new AbortController();
-    setCaptureImages([]);
-    loadCaptureImages({ signal: abortController.signal });
+    if (!isLoading) {
```
nmcharlton, line 61: This doesn't fix the problem – I'm still seeing the same behaviour.
If it did prevent the error, it would result in the wrong captures showing when the page changes during a load.

nmcharlton (changes requested):

nmcharlton (approved):

PR #334, opened 2022-01-28
feat: support country name

dadiorchen:
## Description

Need to add a new API env variable to set the query API URL. 
Assuming that the capture API (for the capture pic on the left) would return with `lat` and `lon` properties.

image

**Issue(s) addressed**

- Resolves #_[insert ID of the issue addressed]_

- [ ] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/components/CaptureMatching/CaptureImage.js
+++ b/src/components/CaptureMatching/CaptureImage.js
@@ -25,7 +25,7 @@ function Country({ lat, lon }) {
         if (res.ok) {
           return res.json();
         } else if (res.status === 404) {
-          setContent('Can not find country at this place');
+          setContent(`Can not find country at lat:${lat}, lon:${lon}`);
           return Promise.reject();
         } else {
           setContent('Unknown error');


.env.test
```diff
@@ -1,2 +1,2 @@
 REACT_APP_WEBMAP_DOMAIN=http://test.treetracker.org
-
+REACT_APP_QUERY_API_ROOT=https://dev-k8s.treetracker.org/query
```
dadiorchen, line 2: Temporarily use dev env for the time being, I will deploy the query API to test env in the near future.

src/components/CaptureMatching/CaptureImage.js
```diff
@@ -10,6 +10,40 @@ import SkipNextIcon from '@material-ui/icons/SkipNext';
...
+  React.useEffect(() => {
+    setContent('loading...');
+    fetch(
+      `${process.env.REACT_APP_QUERY_API_ROOT}/countries?lat=${lat}&lon=${lon}`
+    )
+      .then((res) => {
+        if (res.ok) {
+          return res.json();
+        } else if (res.status === 404) {
+          setContent('Can not find country at this place');
```
nmcharlton, line 28: Maybe we just fall back to displaying raw lat, lon in this case, or nothing at all.
I don't think this message is very useful to the user.

dadiorchen: But if we can't find country for the coordinates, usually it means something wrong, so we'd better let people know that, and if we just simple show lat/lon here, it's a bit confusion for people, why sometimes it's country name and sometimes it's not. I can make this message more clear: `cannot find country at [lat, lon]`

nmcharlton: That works for me

nmcharlton (approved):

PR #339, opened 2022-01-30
Limit capture-match tool to one un-matched capture at a time

### LINKED ISSUE #317: Capture Matching: Request one capture at a time, and only unmatched captures
We currently request _all_ captures in one go when the Capture Matching tool is opened and step through them in the frontend.
This is not particularly scalable. Instead, we should request one capture at a time using server-side pagination as the user steps through the captures.

This means requesting the count separately.

We also need to filter the requests to unmatched captures only (those with `tree_associated=false`).

gwynndp:
## Description

**Issue(s) addressed**

- Resolves #317

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

The request for captures received 1000 captures from the server at once when the UX is that they can only see one capture at a time.

**What is the new behavior?**
The request now has a limit and an offset to make sure that only the first capture is returned from the batch.  It also includes a count of the total number of captures available.

## Breaking change

- [x] No

## Other useful information
This PR depends on changes to the API at this PR: #[52] https://github.com/Greenstand/treetracker-api/pull/52

### DIFF
--- a/src/api/treeTrackerApi.js
+++ b/src/api/treeTrackerApi.js
@@ -1,6 +1,9 @@
 import { handleResponse, handleError, getOrganization } from './apiUtils';
 import { session } from '../models/auth';
 
+// Set API as a variable
+const CAPTURE_MATCH_API = `${process.env.REACT_APP_TREETRACKER_API_ROOT}`;
+
 const CAPTURE_FIELDS = {
   uuid: true,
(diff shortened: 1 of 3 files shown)


nmcharlton (approved):
Nice work, this is really coming together 😄

PR #262, opened 2022-01-01
Stakeholder Tool

gwynndp:
## Description

Make Stakeholder tool more user friendly.

**Issue(s) addressed**
- get unlinked stakeholders based on the current stakeholder detail and the logged in account
- format the Add form: add initial state, format fields, close modal on success
- format Filter form: add initial state, format fields
- optimize linking and unlinking state changes
- add state for count

**What kind of change(s) does this PR introduce?**

- [x] Enhancement

**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## Breaking change

**Does this PR introduce a breaking change?**

- [x] No breaking changes

## Other useful information

- This PR builds upon https://github.com/Greenstand/treetracker-admin-client/pull/200
- corresponds with API PR: https://github.com/Greenstand/treetracker-stakeholder-api/pull/4

PR #340, opened 2022-02-03
Update Stakeholder to use new API routes

gwynndp:
## Description

- update front-end to use the updated API routes (https://github.com/Greenstand/treetracker-stakeholder-api/pull/18)
- remove offset, limit, order, orderBy as filter values since it makes more sense to see all the stakeholders and their children on one page and do sorting on the client
- remove the ability to edit links and add and remove links separate from creating or deleting stakeholders
- allow creating a stakeholder to create the link/relation at the same time
- allow removing a link to also delete a stakeholder
- add dropdowns with values for filters
- add partial text search

**Issue(s) addressed**
- no issue created
- relates to PR for API: https://github.com/Greenstand/treetracker-stakeholder-api/pull/4

- [ ] Enhancement
- [ ] Bug fix
- [x] Refactor

### DIFF
--- a/src/components/Stakeholders/Filter.js
+++ b/src/components/Stakeholders/Filter.js
@@ -15,6 +15,7 @@ import FilterModel from '../../models/FilterStakeholder';
 import { ORGANIZATION_NOT_SET } from '../../models/FilterStakeholder';
 import { AppContext } from '../../context/AppContext';
 import { StakeholdersContext } from '../../context/StakeholdersContext';
+import { localeSort } from '../../common/utils';
 
 const useStyles = makeStyles({
   root: {
@@ -60,14 +61,23 @@ function StakeholderFilter() {
       org.website && websites.add(org.website);
       org.map && maps.add(org.map);
     });
+    // setFilters({
+    //   orgNames: Array.from(orgNames).sort(),
+    //   firstNames: Array.from(firstNames).sort(),
+    //   lastNames: Array.from(lastNames).sort(),
+    //   emails: Array.from(emails).sort(),
+    //   phones: Array.from(phones).sort(),
+    //   websites: Array.from(websites).sort(),
+    //   maps: Array.from(maps).sort(),
+    // });
     setFilters({
-      orgNames: Array.from(orgNames).sort(),
-      firstNames: Array.from(firstNames).sort(),
-      lastNames: Array.from(lastNames).sort(),
-      emails: Array.from(emails).sort(),
-      phones: Array.from(phones).sort(),
-      websites: Array.from(websites).sort(),
-      maps: Array.from(maps).sort(),
+      orgNames: localeSort(Array.from(orgNames)),
+      firstNames: localeSort(Array.from(firstNames)),
+      lastNames: localeSort(Array.from(lastNames)),
+      emails: localeSort(Array.from(emails)),
+      phones: localeSort(Array.from(phones)),
+      websites: localeSort(Array.from(websites)),
+      maps: localeSort(Array.from(maps)),
     });
   }, [stakeholders?.length]);
 
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -9,3 +9,17 @@ export const getVerificationStatus = (active, approved) => {
     return verificationStates.REJECTED;
   }
 };
+
+export const localeSort = (arr, order) => {
+  return arr.sort((a, b) => {
+    const orderVal = order === 'desc' ? -1 : 1;
+    let sortVal = 0;
+    if (a && b) {
+      sortVal = a.localeCompare(b);
+    } else {
+      // if one of them is null or undefined, sort it to the end
+      sortVal = (a || '') > (b || '') ? 1 : -1;
+    }
+    return sortVal * orderVal;
+  });
+};


.env.development
```diff
@@ -5,7 +5,10 @@ REACT_APP_MESSAGING_ROOT=https://dev-k8s.treetracker.org/messaging
 REACT_APP_QUERY_API_ROOT=https://dev-k8s.treetracker.org/query
 REACT_APP_REPORTING_API_ROOT=https://dev-k8s.treetracker.org/reporting
 REACT_APP_TREETRACKER_API_ROOT=https://dev-k8s.treetracker.org/treetracker
+REACT_APP_API_URL=https://dev-k8s.treetracker.org
+REACT_APP_EARNINGS_API_MAPPING=/earnings
 REACT_APP_ENABLE_CAPTURE_MATCHING=true
 REACT_APP_ENABLE_EARNINGS=true
 REACT_APP_ENABLE_PAYMENTS=true
 REACT_APP_ENABLE_MESSAGING=true
+REACT_APP_ENABLE_STAKEHOLDERS=true
```
kparikh9, line 14: I would suggest adding `REACT_APP_STAKEHOLDER_API_ROOT=https://dev-k8s.treetracker.org/stakeholder/stakeholders` to `.env.development`. It might save time for people who are looking to build on top of your changes locally (it took me a while to realize the microservice lived at `/stakeholder/stakeholders` and not `/stakeholders` or `/stakeholder`. Looks like Nick mentioned this in another [PR comment](https://github.com/Greenstand/treetracker-admin-client/pull/262#discussion_r785471670)

src/api/stakeholders.js
```diff
@@ -0,0 +1,189 @@
...
+async function fetchJSON(query, options) {
+  return fetch(query, options).then(handleResponse).catch(handleError);
+}
+
+export default {
+  getStakeholders(id, { offset, rowsPerPage, orderBy, order, filter }) {
+    const orgId = id || getOrganizationId();
+    const filterObj = {
+      where: filter,
+      order: [`${orderBy} ${order}`],
```
kparikh9, line 16: Right now, the `order` and `orderBy` params aren't being used by the stakeholder API, only by the Material table. As a user, if I were to sort, let's say alphabetically, I'd want to see all filtered rows sorted, not just the rows that are visible on the page. If sorting is only expected to order the rows that are returned from the paginated query (thus done in the client), then you can take out `order` from the `filterObj` (since the stakeholder API will always sort by ascending `org_name`. If sorting should be done on the backend, then updates would need to be made to the `stakeholder api` to utilize `options.order`. `handleSort()` in `Stakeholder/Table.js` should also setPage(0), similar to what is done when a filter is applied, to refresh the table.

src/api/stakeholders.js
```diff
@@ -0,0 +1,189 @@
...
+        Authorization: session.token,
+      },
+    };
+
+    return fetchJSON(query, options);
+  },
+
+  async getUnlinkedStakeholders(id, abortController) {
+    const orgId = getOrganizationId();
+    let query = `${STAKEHOLDER_API}/relations`;
```
kparikh9, line 42: The `/relations` endpoint of stakeholder API returns a 500 error when clicking on a stakeholder - `column "owner_id" does not exist`. Maybe this is because the deployed version of the API not the latest?

src/components/Stakeholders/Table.js
```diff
@@ -0,0 +1,119 @@
...
+          <Table aria-label="collapsible table">
+            <TableHead>
+              <TableRow>
+                {columns.map((col) => (
+                  <TableCell
+                    key={col.value}
+                    onClick={() => handleSort(col.value, !order)}
+                  >
+                    <TableSortLabel
+                      active={col.value === !!orderBy}
```
kparikh9, line 75: As a user, I'd like to understand the data that I'm currently looking at in the table. So if a sort is applied, I'd prefer for the arrow indicator to remain. Changing L75 to `active={col.value === orderBy}` persists the arrow label rather than having it unmount immediately.  This may be useful, especially since sort is retained when changing pages.

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,258 @@
...
+  const closeModal = () => {
+    setData(initialState);
+    setOpen(false);
+  };
+
+  const handleChange = (e) => {
+    setData({ ...data, [e.target.name]: e.target.value });
+  };
+
+  const submit = () => {
```
kparikh9, line 78: When adding a Stakeholder, if a user leaves required entries unanswered and attempts to submit, a Validation Error is thrown in the backend. However, I would prevent a request from being sent, which can be done by the frontend. I see that you have passed `required` flags in some form fields, however I'd suggest looking at [this example](https://github.com/Greenstand/treetracker-admin-client/pull/499/files#diff-217d3caf78535283104fa78750472ee22e659c1f349586b09f61ab6b27191c3bR67) that seems to work, which uses a </form> component. This can help in reducing the number of requests sent to the API and prevent the user from making mistakes in the UI leading to incorrect entries in the DB.

kparikh9 (commented):
Hi Gwynn, great job on this huge PR, there are a lot of features that users will appreciate! This is my first time diving into the admin client so take these initial suggestions with grains of salt! If you'd like to discuss anything let me know!

.env.development
gwynndp: Done!

gwynndp: Done!

src/components/Stakeholders/Table.js
gwynndp: Done!

src/api/stakeholders.js
gwynndp: This is due to an update to the API and database that removed the owner_id column

gwynndp: This is because of an update to the API and the removal of the owner_id column.  We'll need to figure out if it's needed or not and the update.

src/api/stakeholders.js
gwynndp: Thank you!  Good point.  Since we're unlikely to have a ton of stakeholders I think I'll drop the table pagination and querying to sort, so moving the sorting to the table makes sense.

src/api/apiUtils.js
```diff
@@ -17,6 +17,7 @@ export async function handleResponse(response) {
 
 // we should call an error logging service, but
 export function handleError(error) {
+  console.log('handleError', error);
```
nmcharlton, line 20: Use `log.debug` or similar

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+                name="website"
+                onChange={handleChange}
+                value={data.website}
+                onKeyDown={handleEnterPress}
+              />
+            </FormControl>
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <TextField
+                className={classes.textField}
+                label="Logo Url"
```
nmcharlton, line 261: Logo URL

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+            </FormControl>
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <TextField
+                error={!!errors?.phone}
+                helperText={errors.phone}
+                className={classes.textField}
+                label="Phone"
+                variant="outlined"
+                name="phone"
+                type="number"
```
nmcharlton, line 241: type="tel"

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <TextField
+                className={classes.textField}
+                label="Website"
+                variant="outlined"
+                name="website"
+                onChange={handleChange}
+                value={data.website}
+                onKeyDown={handleEnterPress}
+              />
```
nmcharlton, line 256: type="url"

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+              <TextField
+                error={!!errors?.email}
+                helperText={errors.email}
+                className={classes.textField}
+                label="Email"
+                variant="outlined"
+                name="email"
+                onChange={handleChange}
+                value={data.email}
+              />
```
nmcharlton, line 231: type="email"

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <TextField
+                className={classes.textField}
+                label="Logo Url"
+                variant="outlined"
+                name="logo_url"
+                onChange={handleChange}
+                value={data.logo_url}
+                onKeyDown={handleEnterPress}
+              />
```
nmcharlton, line 267: type="url"

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+            </FormControl>
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <TextField
```
nmcharlton, line 270: Perhaps we should try to pre-populate this based on the organization name.

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,354 @@
...
+                <FormControlLabel
+                  label="Child"
+                  value="children"
+                  control={<Radio />}
+                  onChange={handleChange}
+                />
+              </RadioGroup>
+              <TextField
+                className={classes.textField}
+                label="Relation ID"
```
nmcharlton, line 311: Is this meant to be the ID for the related stakeholder? If so, this might be more useable as a dropdown.

src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,342 @@
...
+    setOpen(false);
+  };
+
+  const handleChanges = (e) => {
+    const key = e.target.name === 'organization_id' ? 'id' : e.target.name;
+    const value =
+      e.target.name === 'organization_id'
+        ? orgList.find((o) => o.stakeholder_uuid === e.target.value)
+            ?.stakeholder_uuid || ''
+        : e.target.value !== 'Not Set'
```
nmcharlton, line 85: 'Not Set' may be a little confusing – we use it elsewhere to mean show only records with null/empty values in this field.

nmcharlton:
@gwynndp I've restarted my review of this PR. I'm adding comments as I go, but most can be addressed as issues once the tool is merged. Don't feel obliged to make any changes for now, I'll let you know if there are any showstoppers.

src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,342 @@
...
+      id: ORGANIZATION_NOT_SET,
+      name: 'Not Set',
+      value: '',
+    },
+  ];
+
+  const defaultTypeList = [
+    {
+      name: 'Not set',
+      value: undefined,
```
nmcharlton, line 121: `undefined` isn't handled properly in `api/stakeholders.js` – it only omits filters set to `''`.

src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,342 @@
...
+      org.org_name && orgNames.add(org.org_name);
+      org.first_name && firstNames.add(org.first_name);
+      org.last_name && lastNames.add(org.last_name);
+      org.email && emails.add(org.email);
+      org.phone && phones.add(org.phone);
+      org.website && websites.add(org.website);
+      org.map && maps.add(org.map);
+    });
+    setFilters({
+      orgNames: Array.from(orgNames).sort(),
```
nmcharlton, line 64: Use `localeCompare()` for the sorts – by default, `Array.sort()` uses Unicode values.

src/components/Stakeholders/Table.js
```diff
@@ -0,0 +1,127 @@
...
+    if (col) {
+      sorted = stakeholders.sort((a, b) => {
+        let first = a[col];
+        let second = b[col];
+        if (col === 'name') {
+          first = `${a.org_name} ${a.first_name} ${a.lastname}`;
+          second = `${b.org_name} ${b.first_name} ${b.lastname}`;
+        }
+        return (
+          (first.trim() > second.trim() ? 1 : -1) * (order === 'desc' ? -1 : 1)
```
nmcharlton, line 57: `first` and `second` might be `null` or `undefined` here.

src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,342 @@
...
+      <Button
+        variant="text"
+        color="primary"
+        onClick={() => setOpen(true)}
+        startIcon={<FilterIcon />}
+        className={classes.ml}
+      >
+        Filter
+      </Button>
+      {Object.keys(filter).length > 0 && (
```
nmcharlton, line 154: `filter` always has a complete set of keys with empty string values by default, so this always shows a button.

src/components/Stakeholders/Filter.js
```diff
@@ -0,0 +1,342 @@
...
+        <Button onClick={resetFilters}>Reset Filters</Button>
+      )}
+      <Dialog open={open} onClose={close} fullWidth={true} maxWidth={'md'}>
+        <DialogTitle>Filters</DialogTitle>
+        <DialogContent>
+          <FormControl className={classes.root}>
+            <TextField
+              data-testid="type-dropdown"
```
nmcharlton, line 162: I'm not sure about the use of dropdowns for these filters; they seem a bit odd from a UX point of view given that each list is quite long.
Perhaps `Autocomplete` would be more appropriate.
To be honest, I'm not sure how useful the filters will be, given we have the more powerful and efficient search facility.

src/api/apiUtils.js
gwynndp: Done!

src/components/Stakeholders/Add.js
gwynndp: Ok, good idea

src/components/Stakeholders/Add.js
gwynndp: Would it exist already or would this be a way of naming it?

We should be able to query for it.

src/components/Stakeholders/FilterBar.js
```diff
@@ -0,0 +1,33 @@
...
+
+const useStyles = makeStyles({
+  flex: {
+    display: 'flex',
+    alignItems: 'center',
+  },
+  spaceBetween: {
+    justifyContent: 'space-between',
+  },
+  my: {
```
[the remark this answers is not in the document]
gwynndp: Done!

src/components/Stakeholders/Filter.js
gwynndp: Agreed about Autocomplete being better.  The dropdowns make sense if the user doesn't know the options and if there are enough stakeholders that combining filters would be helpful and informative.

.env.development
```diff
@@ -5,10 +5,12 @@ REACT_APP_GROWER_QUERY_API_ROOT=https://dev-k8s.treetracker.org/grower-account-q
 REACT_APP_IMAGES_API_ROOT=https://dev-k8s.treetracker.org/images
 REACT_APP_MESSAGING_ROOT=https://dev-k8s.treetracker.org/messaging
 REACT_APP_QUERY_API_ROOT=https://dev-k8s.treetracker.org/query
-REACT_APP_REPORTING_API_ROOT=https://dev-k8s.treetracker.org/reporting
 REACT_APP_REGION_API_ROOT=https://dev-k8s.treetracker.org/regions
+REACT_APP_REPORTING_API_ROOT=https://dev-k8s.treetracker.org/reporting
+REACT_APP_STAKEHOLDER_API_ROOT=https://dev-k8s.treetracker.org/stakeholder/stakeholders
```
nmcharlton, line 10: Is this really the root?

src/components/Stakeholders/StakeholderDetail/List.js
```diff
@@ -0,0 +1,65 @@
...
+
+function StakeholderList({ id, data, type, linked }) {
+  const classes = useStyles();
+  const [isLinked, setIsLinked] = useState(linked);
+  const { deleteStakeholder } = useContext(StakeholdersContext);
+
+  const handleChange = (e, data) => {
+    setIsLinked(!isLinked);
+    const confirmed = window.confirm(
+      'Are you sure? Continuing will delete the stakeholder from the system.'
```
nmcharlton, line 25: If we're making this a delete function rather than unlink, let's use a trash icon.
If deletion is just a placeholder for unlinking, can you add a comment to explain what's happening here?

src/components/Stakeholders/StakeholderDetail/List.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+});
+
+function StakeholderList({ id, data, type, linked }) {
+  const classes = useStyles();
+  const [isLinked, setIsLinked] = useState(linked);
+  const { deleteStakeholder } = useContext(StakeholdersContext);
+
+  const handleChange = (e, data) => {
+    setIsLinked(!isLinked);
```
nmcharlton, line 23: I don't expect linked to be a state in this way. The `unlink` icon signifies an action to unlink the item from the focussed stakeholder (i.e., delete the relationship). It will always be an `unlink` icon.

src/components/Stakeholders/Add.js
nmcharlton: Good point. I thought of this as a way of naming the map, but you're right that it might already exist.
This might be a good way to add new maps via the webmap API, but I'm not sure how automated this process currently is.
Leave it as it is for now and we can raise an issue to discuss options.

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,357 @@
...
+            <FormControl xs={12} sm={6} className={classes.root}>
+              <FormLabel id="relation">Relationship</FormLabel>
+              <RadioGroup
+                aria-labelledby="radio-buttons-group-label"
+                name="relation"
+                value={data.relation}
+                className={classes.radioGroup}
+                onKeyDown={handleEnterPress}
+              >
+                <FormControlLabel
```
nmcharlton, line 295: Is there a reason the two radio options are formatted differently?

src/components/Stakeholders/Add.js
```diff
@@ -0,0 +1,357 @@
...
+                  control={
+                    <Radio
+                      label="Parent"
+                      value="parents"
+                      onChange={handleChange}
+                    />
+                  }
+                  label="Parent"
```
nmcharlton, line 303: It's not clear whether _Parent_ means that this stakeholder is the parent or the ID entered here is the parent.
It'd be clearer to replace this with an option to _Link to parent_ that presents a list of stakeholders to link this one to.
Realistically, I only see us linking new stakeholders to their parents at the point of creation.

src/components/Stakeholders/StakeholderDetail/StakeholderDetail.js
```diff
@@ -0,0 +1,143 @@
...
+  };
+
+  const [open, setOpen] = useState(false);
+  const classes = useStyles();
+
+  const openModal = () => {
+    setOpen(true);
+  };
+
+  const closeModal = () => {
```
nmcharlton, line 78: If `isEditing` is true, show a warning before closing to prevent users losing unsaved changes.

src/components/Stakeholders/Table.js
```diff
@@ -0,0 +1,133 @@
...
+      ) : (
+        <TableContainer component={Paper}>
+          <Table aria-label="collapsible table">
+            <TableHead>
+              <TableRow>
+                {columns.map((col) => (
+                  <TableCell
+                    key={col.value}
+                    onClick={() =>
+                      handleSort(col.value, order === 'desc' ? 'asc' : 'desc')
```
nmcharlton, line 94: If this column is not currently selected for sort, default to `asc`.
Else, toggle between `asc` and `desc` as is implemented here.

The current behaviour feels very odd.

nmcharlton (approved):
Nice work!
I've completed a functional review, but am happy for us to merge and treat all outstanding comments as issues.

.env.development
gwynndp: Yes :)

src/components/Stakeholders/Add.js
gwynndp: Are you referring to a radio option on a different page?

nmcharlton: These two `FormControlLabel`s have different syntax:
```
                <FormControlLabel
                  control={
                    <Radio
                      label="Parent"
                      value="parents"
                      onChange={handleChange}
                    />
                  }
                  label="Parent"
                />
                <FormControlLabel
                  label="Child"
                  value="children"
                  control={<Radio />}
                  onChange={handleChange}
                />
```
The result is the same (I assume), but they should be consistent.

PR #358, opened 2022-02-04
fix(session): check for valid session before API calls

### LINKED ISSUE #290: 401 Unauthorized response to check_session doesn't trigger logout
When a user's session has expired, the JWT token changes or a developer switches between a local backend and a cloud backend, reloading the dashboard triggers several API calls that return 401 (unauthorized) errors.

This triggers confusing errors in development:
Screenshot 2022-01-14 at 21 09 52
In production, the dashboard appears stuck, with no indication of what's happened:
Screenshot 2022-01-14 at 21 10 12
It should be up to `/auth/check_session` to check whether the session is still valid, and a 401 response to this should trigger a logout redirect, but this doesn't seem to be working properly.
Other API calls should ideally be paused until `check_session` has completed.

popietree:
## Description

Add condition session.token to execute checkSession only if session.token and other conditions are true. Otherwise 5 more API calls from getTotalUnprocessed, getTotalVerified, loadOrganizations, getTotal, and getTotalGrowerCount will fetch data. 

**Issue(s) addressed**
- Resolves #290 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

The app does not logout properly upon 401 error. checkSession is not checking for valid session so logging in will trigger all API calls from mounting DashStat Components.  

**What is the new behavior?**
App will logout and stop making API calls when JWT token changes. 

## Breaking change

No

## Other useful information
Currently no catch method on checkSession to logout on error. Does not handle 401 error specifically

### DIFF
--- a/src/context/AppContext.js
+++ b/src/context/AppContext.js
@@ -205,11 +205,11 @@ export const AppProvider = (props) => {
   function checkSession() {
     const localToken = JSON.parse(localStorage.getItem('token'));
     const localUser = JSON.parse(localStorage.getItem('user'));
-    if (localToken && localUser) {
+
+    if (localToken && localUser && session.token) {
       // Temporarily log in with the localStorage credentials while
       // we check that the session is still valid
       login(localUser, localToken);
-
       axios
         .get(
           `${process.env.REACT_APP_API_ROOT}/auth/check_session?id=${localUser.id}`,
@@ -232,6 +232,7 @@ export const AppProvider = (props) => {
             logout();
           }
         });
+
       return true;
     }
     return false;


popietree:
Needs review

src/context/AppContext.js
```diff
@@ -205,11 +205,11 @@ export const AppProvider = (props) => {
   function checkSession() {
     const localToken = JSON.parse(localStorage.getItem('token'));
     const localUser = JSON.parse(localStorage.getItem('user'));
-    if (localToken && localUser) {
+
+    if (localToken && localUser && session.token) {
```
nmcharlton, line 209: Unfortunately, this doesn't fix the underlying problem, and has the undesirable side effect of logging the user out every time they refresh the page.

We _should_ call `/auth/check_session` if there's a local token and local user, but should ideally suspend other API calls until this check has finished and the user has been logged out (if the response is 401).

nmcharlton (changes requested):

nmcharlton:
This will likely be impacted by the introduction of Keycloak, so we'll pause development for now.

Issue #367, opened 2022-02-06
Capture Matching: Use OptimizedImage component in place of raw images


nmcharlton:
Throughout the Admin Panel we use the `OptimizedImage` component, a wrapper to the statically.io image resize and cache service, to ensure images load efficiently. The Capture Matching tool currently uses raw image URLs, so can be slow to load.

It should instead use `OptimizedImage` for the unmatched captures and candidate match images.

PR #364, opened 2022-02-05
fix: assorted improvements to capture matching UI

nmcharlton:
## Description

This PR introduces several improvements to the Capture Matching tool:
1. Show date and proximity of candidate match captures
2. Display grower image and name
3. Tidy styling of various elements, including buttons and general spacing
4. Make unmatched capture image fully visible without scrolling
5. Relabel _Same / DIfferent Tree_ buttons as _Confirm Match_ and _Dismiss_ to make their actions clearer
6. Animate collapse of candidate match card
7. Remove unused filter placeholder chips

Screenshot 2022-02-05 at 21 37 36

https://user-images.githubusercontent.com/5558838/152700065-03078205-98a5-4e89-b41d-b2d07f111714.mov

**Issue(s) addressed**

- Resolves #361 
- Resolves #362 
- Resolves #363 

**What kind of change(s) does this PR introduce?**

- [x] Enhancement
- [x] Bug fix
- [ ] Refactor

**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #370, opened 2022-02-06
Optimized capture image

gwynndp:
## Description

Update CaptureImage to use OptimizedImage instead of a plain  so that images are loaded faster.

**Issue(s) addressed**

- Resolves #_[367]_https://github.com/Greenstand/treetracker-admin-client/issues/367

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/CaptureMatching/CaptureImage.js
+++ b/src/components/CaptureMatching/CaptureImage.js
@@ -64,9 +64,9 @@ const useStyles = makeStyles((theme) => ({
     justifyContent: 'center',
     padding: theme.spacing(2),
   },
-
   imgContainer: {
-    objectFit: 'contain',
+    position: 'relative',
+    flexGrow: 1,
   },
   captureInfo: {
     display: 'flex',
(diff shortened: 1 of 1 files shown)


nmcharlton:
@gwynndp Do you think you could apply this change to #364?
I made some changes in the area and we're likely to end up with merge issues.

nmcharlton:
There seems to be a problem with statically.io and rotation at the moment. I'm not sure what's causing it, but when our images are served raw, the orientation is correct. When they're served through Statically, they're not rotated.

Raw:
Screenshot 2022-02-07 at 20 19 09
Statically:
Screenshot 2022-02-07 at 20 19 12

nmcharlton:
Statically is returning a png instead of a jpg, so the orientation data is being lost.

ZavenArra (approved):

gwynndp:
Maybe we should revert this merge until we can figure out the orientation data @ZavenArra 
@nmcharlton Do you think we could load the optimized image and then replace it with the raw image when it loads?

nmcharlton:
> Maybe we should revert this merge until we can figure out the orientation data @ZavenArra 
> 
> @nmcharlton Do you think we could load the optimized image and then replace it with the raw image when it loads?

I agree that we should probably revert. As for loading both, I'm less keen as this will cause a jump in orientation when the raw image loads.

gwynndp:
Status update: 
I can't reproduce this rotation issue but I have looked into a few different libraries that might help us read the EXIF orientation data.  I suspect that if this is an issue we might have to handle it on the server side to make sure the orientation is being stored in S3.  From tests today I'm not getting ANY data from the images stored there but without being able to reproduce the issue in the browser I can't be sure if there is really no data or if the libraries I'm trying aren't working in my tests.

PR #438, opened 2022-02-24
Fix bug with DistanceTo

gwynndp:
## Description

Capture match wasn't loading because of a bug with DistanceTo.

It was referencing .lat and .lon when the properties are now .latitude and .longitude
And the values were strings when the calculation needed floats

**Issue(s) addressed**
- no issue

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

NOTE:
- You will see there are some distances showing  as "NaNm away".  This is because the candidateCapture data doesn't have values for it's latitude and longitude

https://monosnap.com/file/1AKn9qsdiK3Z9J48eRgqsuv3KsCNMe

### DIFF
--- a/src/components/CaptureMatching/CandidateImages.js
+++ b/src/components/CaptureMatching/CandidateImages.js
@@ -114,10 +114,10 @@ const useStyles = makeStyles({
 function DistanceTo({ lat1, lon1, lat2, lon2 }) {
   const [content, setContent] = useState('');
   if (
-    lat1 === 'undefined' ||
-    lon1 === 'undefined' ||
-    lat2 === 'undefined' ||
-    lon2 === 'undefined'
+    lat1 === undefined ||
+    lon1 === undefined ||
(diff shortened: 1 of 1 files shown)


nmcharlton (changes requested):
From my testing it looked like the missing `latitude` and `longitude` values come from the candidate _tree_. If it has no captures, we use the tree itself as a capture (`lat`,`lon`,`created_at`,`image_url`).

We need the API to be consistent, or to always populate the captures for candidate trees.

Also, I'd suggest fixing the checks for `lat1 === 'undefined'` etc. to be `lat1 === undefined` so we catch those instances and don't display NaN (which is invariably confusing for users).

ZavenArra (approved):

gwynndp:
@nmcharlton Thanks, I missed that the properties were named differently.  See if this works for the moment.

nmcharlton (approved):
Looks good

PR #454, opened 2022-03-02
Messaging refactored to work with updated Messaging API

gwynndp:
## Description

Update client to work with new Messaging API changes

**Issue(s) addressed**

- no issue

- [ ] Enhancement
- [ ] Bug fix
- [x] Refactor

### DIFF
--- a/src/components/Messaging/MessageBody.js
+++ b/src/components/Messaging/MessageBody.js
@@ -356,9 +356,11 @@ const SenderInformation = ({ message, messageRecipient, type, id }) => {
             <Typography>
               <b>DATE:</b> {dateFormat(message?.composed_at, 'yyyy/mm/dd')}
             </Typography>
-            <Typography>
-              <b>TO:</b> {message?.bulk_message_recipients[0].recipient}
-            </Typography>
+            {message?.bulk_message_recipients && (
+              <Typography>
+                <b>TO:</b> {message?.bulk_message_recipients[0]?.recipient}
+              </Typography>
+            )}
           </>
         )}
 


.env.development
```diff
@@ -1,7 +1,7 @@
 REACT_APP_WEBMAP_DOMAIN=http://dev.treetracker.org
 REACT_APP_API_ROOT=https://dev-k8s.treetracker.org/api/admin
 REACT_APP_EARNINGS_ROOT=https://dev-k8s.treetracker.org/earnings
-REACT_APP_MESSAGING_ROOT=https://dev-k8s.treetracker.org/messaging
+REACT_APP_MESSAGING_ROOT=http://localhost:3004
 REACT_APP_QUERY_API_ROOT=https://dev-k8s.treetracker.org/query
```
ZavenArra, line 5: Don't change the defaults here!  You can run locally by making your own local .env file and starting with NODE_ENV= that environment.

src/api/messaging.js
```diff
@@ -54,7 +54,7 @@ export default {
       .then(handleResponse)
       .catch(handleError);
   },
-  getMessage(author_handle) {
+  getMessages(author_handle) {
     const query = `${process.env.REACT_APP_MESSAGING_ROOT}/message?author_handle=${author_handle}`;
```
ZavenArra, line 59: _author_handle_ has been changed to just _handle_

src/components/Messaging/MessageBody.js
```diff
@@ -333,22 +364,23 @@ const MessageBody = ({ messages, messageRecipient }) => {
...
-      if (res) {
-        setRecipientId(res.id);
-      }
+    const res = authors.find((author) => author.handle === messageRecipient);
+    if (res?.id) {
+      setRecipientId(res.id);
+    } else {
+      // if not an author it must be an org or a region and the id is the handle
+      setRecipientId(messageRecipient);
     }
```
ZavenArra, line 384: If the recipient is an org or a region, there is no recipient id shows to the sender of that message.  Rather this is populated by the bulk_message_recipients property, which an an array describing the recipient parameters.

src/components/Messaging/Messaging.js
```diff
@@ -85,8 +91,9 @@ const Messaging = () => {
   }, [messages]);
 
   const handleListItemClick = (e, i, userName) => {
-    setSelectedIndex(i);
+    console.log('handleListItemClick', i, userName);
     setMessageRecipient(userName);
```
ZavenArra, line 95: Can we start using loglevel in the admin panel instead of console.log?

src/context/MessagingContext.js
```diff
@@ -34,26 +34,30 @@ export const MessagingProvider = (props) => {
...
+        } else if (message.type === 'survey') {
           let key = message.survey.id;
           if (!grouped[key]) {
             grouped[key] = [];
           }
           grouped[key].push(message);
+        } else if (message.type === 'announce') {
+          // assume it's an announcement
+          let key = message.subject;
+          if (grouped[key]) {
```
ZavenArra, line 54: One issue with grouping this way is that multiple messages with the same announcement title will get grouped together.  In fact, an announce message is always a group with just one single message - the announcement that gets sent.  All responses are in the individual user channels.  So we don't need to group when we have an announce message - it's a group of 1.

src/context/MessagingContext.js
```diff
@@ -34,26 +34,30 @@ export const MessagingProvider = (props) => {
...
+            const date = Date.now();
+            console.log('date', date);
+            grouped[`${key}-${Date.now()}`] = [message];
+          } else {
```
ZavenArra, line 58: hmm ok i see, maybe that group key using Date.now() is handling the grouping logic I mentioned above?

ZavenArra (changes requested):
Looks great.  Just a few small items to address in my comments.

.env.development
gwynndp: Nick caught me on the same -- Will do from here on

src/context/MessagingContext.js
gwynndp: I handled that by adding the date to the title if there was already a match

src/components/Messaging/MessageBody.js
gwynndp: Wouldn't it be useful for the sender to see who they sent that message to, if they have more than one option to target? which region? which org?

src/context/MessagingContext.js
gwynndp: If the bulk messages have ids then we can group them by that unique id and not worry about adding the date to the name.  Surveys have the id, but I don't think the announce messages do.

It might be interesting to group all the announcements by organization/region id too...

ZavenArra (approved):

PR #455, opened 2022-03-02
Messaging: sort recipient handles, add views for loading and for 'no messages'

### LINKED ISSUE #429: When no messages are present, show appropriate UI
When new no messages are present for an admin panel user, the UI shows as below:
image

Instead, we should show an empty right panel with the text 'You have no messages'

gwynndp:
## Description

- add loading view
- add view for no messages
- alphabetize recipient handles in list when creating a new message

**Issue(s) addressed**

- Resolves #429 
- For #395, it already allows typing and filtering but the handles weren't alphabetized so I added that

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/api/apiUtils.js
+++ b/src/api/apiUtils.js
@@ -6,19 +6,17 @@ export async function handleResponse(response) {
   if (response.ok) return response.json();
 
   // server-side validation error occurred.
-  // Server side validation returns a string error message, so parse as text instead of json.
-  const error = await response.text();
+  const error = await response.json();
   log.debug('handleResponse error ---', error);
 
-  if (response.status === 400) {
-    throw new Error(error);
-  }
-  if (response.status === 500 && error.includes('Author handle not found')) {
-    // server-side error occurred. Author account not found.
+  if (response.status === 404 && error.message === 'Author handle not found') {
     return {
       error: true,
-      message: error,
+      message: error.message,
     };
+  }
+  if (response.status === 400) {
+    throw new Error(error);
   }
   throw new Error('Network response was not ok.');
 }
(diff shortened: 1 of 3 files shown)


.env.development
```diff
@@ -7,5 +7,6 @@ REACT_APP_REPORTING_API_ROOT=https://dev-k8s.treetracker.org/reporting
 REACT_APP_TREETRACKER_API_ROOT=https://dev-k8s.treetracker.org/treetracker
 REACT_APP_ENABLE_CAPTURE_MATCHING=true
 REACT_APP_ENABLE_EARNINGS=true
-REACT_APP_ENABLE_PAYMENTS=true
 REACT_APP_ENABLE_MESSAGING=true
+REACT_APP_ENABLE_PAYMENTS=true
+REACT_APP_REPORTING_ENABLED=false
```
ZavenArra, line 12: Why is REPORTING marked as false here?  I believe reporting should always be true now, as it is deployed into production.

src/api/apiUtils.js
```diff
@@ -3,12 +3,24 @@ import { session } from '../models/auth';
...
+  console.log('handleResponse error ---', error);
+
   if (response.status === 400) {
-    // server-side validation error occurred.
-    // Server side validation returns a string error message, so parse as text instead of json.
-    const error = await response.text();
     throw new Error(error);
   }
+  if (response.status === 500 && error.includes('author')) {
+    // server-side error occurred. Author account not found.
```
ZavenArra, line 16: is checking for the string 'author' enough to guarantee there is only one candidate error that could cause this 500 ?

ZavenArra (changes requested):
Merged another PR and created conflicts in MessageBody.js. Resolve this and a couple small comments I left, and should be good to merge!

src/api/apiUtils.js
gwynndp: Probably not specific enough, but I reasoned that it would either be that their admin handle didn't exist or some other error that we haven't encountered already when requesting authors that they should talk to us about.

I'll adjust it to the specific SQL error for now and we can worry about other error-handling later

gwynndp: Ah! I see the message has been updated.  Before I was just getting the SQL error

PR #483, opened 2022-03-11
filter author by org

ZavenArra:
- feat: use grower account query service to get authors
- wip: get the organization uuid from the policy and pass that to the API request

## Description

**Issue(s) addressed**

- Resolves #_[insert ID of the issue addressed]_

- [ ] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/api/messaging.js
+++ b/src/api/messaging.js
@@ -37,19 +37,7 @@ export default {
     console.log(mappedRes);
     return mappedRes;
   },
-  // getAuthorAvatar(handle) {
-  //   const query = `${process.env.REACT_APP_TREETRACKER_API_ROOT}/grower_accounts?wallet=${handle}`;
 
-  //   return fetch(query, {
-  //     method: 'GET',
-  //     headers: {
-  //       'content-type': 'application/json',
-  //       Authorization: session.token,
-  //     },
-  //   })
-  //     .then(handleResponse)
-  //     .catch(handleError);
-  // },
   postRegion(payload) {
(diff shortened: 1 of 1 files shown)


src/context/MessagingContext.js
```diff
@@ -112,8 +112,8 @@ export const MessagingProvider = (props) => {
     setIsLoading(false);
   };
 
-  const loadAuthors = async () => {
-    const res = await api.getAuthors();
+  const loadAuthors = async (organizationId) => {
+    const res = await api.getAuthors(organizationId);
 
     if (res.authors) {
       let result = res.authors.filter(
```
gwynndp, line 119: The res doesn't have an authors property.  It has a growerAccounts property instead:

`const loadAuthors = async (organizationId) => {
    const res = await api.getAuthors(organizationId);

    if (res.growerAccounts) {
      let result = res.growerAccounts.filter(
        (author) => author.handle !== user.userName
      );
      setAuthors(result);
    }
  };`
  
  
  Otherwise, Nice Work! @ZavenArra

ZavenArra: ah got it

ZavenArra:
I think that 'res' should not leave the context of the API level, right?  If the call is getAuthors, then we should just return the authors array itself.  I see that other api functions not implemented this way either, but I think they probably should be.

If I call a function 'getCount', which I should back is the count.  All http and payload structure knowledge should be contained in the API layer, which then should then hand back objects or collections of object in the domain language of the application.

PR #502, opened 2022-03-14
Responsive display of captures verify tool

### LINKED ISSUE #234: Improve responsive display of captures in Verify tool
# Current Behaviour
The Verify tool currently has some responsiveness built in for displaying the capture grid differently on different resolution screens.
The problem is that the scaling is not very consistent and the maximum captures per row is four.

On large/high-resolution screens, the user sees four very large images per row and the proportional increase in capture height means there are typically only ever two rows fully visible.
This is inefficient when batch processing captures.

## 800px

image

## 1440px

image

## 1920px

image

## 2560px

image

# Desired Behaviour

Capture card/image size should stay within a sensible range and the number of captures per per row should not be limited.
We should also add an option to toggle the view between large and small captures.
This would allow the user to select the appropriate view for the job in hand - verification and species tagging often requires large capture images for close inspection, while batch tagging for audit purposes can be done with smaller images.

## Proposed Sizes
**Small:** 200px-250px
**Large:** 320px-400px

bakozetra:
## Description

Improved responsive display of captures in Verify tool.

**Issue(s) addressed**

- Resolves #234

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

 Scaling is not very consistent and the maximum captures per row are four.

**What is the new behavior?**
- Adding a toggle button that allows the user to click the large or small images.
- Fixing the responsiveness.


nmcharlton (commented):
We shouldn't delete package-lock.json. We need to maintain it in the project.

src/components/Verify.js
```diff
@@ -322,10 +340,18 @@ const Verify = (props) => {
...
 
   const captureImageItems = captureImages
+
     .concat(placeholderImages)
     .map((capture) => {
       return (
-        <Grid item xs={12} sm={6} md={4} xl={3} key={capture.id}>
+        <Grid
+          item
+          xs={showBigSize ? 0 : 12}
```
nmcharlton, line 349: I suspect responsive grid sizing is too limiting here. It only adjusts to the viewport width, so doesn't account for the fixed-width tagging panel.
I suggest doing your own calculation for items per row based on the container width and setting the item width to `calc(100% / ${numCaptures}`.

See https://gist.github.com/nmcharlton/0bee3372c2eec090ee5b0d97342e0c11 for a possible solution.

bakozetra:
> Hi @bakozetra. Are you able to try to address the comments on this PR?

Hi @nmcharlton, Sorry for the late reply, I missed the message from Github. I was trying to apply the changes but I could not finish the task as I am busy with work now. I am pushing for all the changes and I am happy if you take it over. I don't think I will have time to finish this in near future.
>

nmcharlton:
No problem @bakozetra. Thanks for all your efforts on this.

nmcharlton:
Superseded by #681

PR #529, opened 2022-03-16
fix: do not loose side panel when error on bulk message send

ZavenArra:


ZavenArra:
this PR is failing CI on npm run lint, but this same command succeeds locally.  What is the issue?

Issue #473, opened 2022-03-05
Handling API query errors gracefully and inform the user of failures


nmcharlton:
We've had several instances recently of errors in API responses not being handled gracefully and not being reported to the user.
In several places, we have no error handling at all, so either quietly ignore errors or assume success, often resulting in unexpected behaviour.

Ideally, we would have a mechanism in the Admin Panel to allow users to report errors directly when they occur, along with the network request and response details, but a good start would be graceful handling of failures.

nmcharlton:
@xxGreyscale That would be awesome. I think a UI consultation would be good, but it doesn’t need to be anything too fancy 😀

nmcharlton:
Hi @xxGreyscale. Have you been able to make any progress on this issue? The team has been discussing the best strategy for API error handling across the Admin Panel, and @gwynndp has proposed some changes in #566, focussed on the Messaging tool. It would be good to share and align the work on this area.

xxGreyscale:
So far I haven't made any significant progress ready to be aligned with the main repository. Just figured out a way of handling most errors. But yet to be implemented

OlhaD:
@nmcharlton I would like to work on it.

PR #566, opened 2022-03-22
Improve error message for admin user not configured

### LINKED ISSUE #564: Improve error message for admin user not configured
If an admin panel user is not yet configured, the initial call to get messages for that users returns an error.   'Configured' here means that there is an entry in the author with a handle matching the username of the admin panel user.

Currently, we report this as a 'network error', but it's not an issue with the network, rather it's a configuration error.  The user is not yet configured, but can easily be configured by the tech team if approved.   We should change this error message to say

"Configuration Error:  Your user is not yet configured for messaging access.  Please contract technical support"

gwynndp:
## Description

- update the error message
- make sure all user errors are passed along so they can be handled

**Issue(s) addressed**

- Resolves #564

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/Messaging/Survey.js
+++ b/src/components/Messaging/Survey.js
@@ -8,7 +8,6 @@ import {
   Button,
   IconButton,
   FormControl,
-  FormHelperText,
 } from '@material-ui/core';
 import { Autocomplete } from '@material-ui/lab';
 import { Close } from '@material-ui/icons';
@@ -297,10 +296,6 @@ const SurveyForm = ({ setToggleSurvey }) => {
             value={`question${num}`['prompt']}
             onChange={(e) => handleChange(e, `question${num}`)}
             {...(errors?.questions && { error: true })}
-            // {...(errors?.questions && {
-            //   error: true,
-            //   helperText: errors.questions,
-            // })}
           />
           <GSInputLabel text={`Question ${num} Answer Options`} />
           <TextField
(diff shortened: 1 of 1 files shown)


src/api/apiUtils.js
```diff
@@ -9,15 +9,15 @@ export async function handleResponse(response) {
   const error = await response.json();
   log.debug('handleResponse error ---', response.status, error);
 
-  if (response.status === 422) {
+  // pass along user errors
```
nmcharlton, line 12: I'm a little concerned about splitting out error handling into 4xx and 5xx like this.
It puts the onus on the caller to implement two separate ways of catching errors.
A 4xx doesn't necessarily mean a _user_ error, it could just be caused by a bug in the client or a mismatch with the API.

I think my preference would be to always throw the error object (status and message) and expect the caller to wrap the call in a `try...catch`, but I'm happy to be challenged.
If the caller knows what to do with the status code, it can behave accordingly. Otherwise, it rethrows and a higher-level `catch` can be responsible for showing an alert when something truly unhandled happens.
What do you think?

nmcharlton: @xxGreyscale may have an opinion on this with the work on #473.

gwynndp: I'll give it a go!

nmcharlton (approved):

PR #625, opened 2022-04-03
Create a separate component for selecting organizations

### LINKED ISSUE #379: Filters - reuse and create components from common fields

gwynndp:
## Description

- Create a separate component for selecting organizations
- allow different handling of the organization by different filters
- Update filters to use the new component

**Issue(s) addressed**

- Resolves #379

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/tests/verify.test.js
+++ b/src/components/tests/verify.test.js
@@ -169,10 +169,9 @@ describe('Verify', () => {
       userEvent.selectOptions(orglist, orgSelected);
 
       userEvent.click(screen.getByText(/apply/i));
-      // screen.logTestingPlaygroundURL();
-      expect(screen.getByRole('button', { name: /filter 2/i })).toBeTruthy();
+      expect(screen.getByRole('button', { name: /filter 1/i })).toBeTruthy();
       //this function is still returning 1
-      // expect(verifyValues.filter.countAppliedFilters()).toBe(2);
+      expect(verifyValues.filter.countAppliedFilters()).toBe(1);
     });
 
     // it('renders side panel', () => {
@@ -181,7 +180,6 @@ describe('Verify', () => {
     // });
 
     it('renders captures gallery', () => {
-      // screen.logTestingPlaygroundURL();
       const pageSize = screen.getAllByText(/captures per page:/i);
       expect(pageSize).toHaveLength(2);
 
@@ -194,7 +192,6 @@ describe('Verify', () => {
       });
       expect(captureDetails).toHaveLength(4);
       userEvent.click(captureDetails[0]);
-      // screen.logTestingPlaygroundURL();
       expect(screen.getByText(/capture data/i)).toBeInTheDocument();
       expect(screen.getByText(/grower identifier/i)).toBeInTheDocument();
       expect(screen.getByText(/grower1@some.place/i)).toBeInTheDocument();
@@ -210,7 +207,7 @@ describe('Verify', () => {
       });
       expect(growerDetails).toHaveLength(4);
       userEvent.click(growerDetails[0]);
-      screen.logTestingPlaygroundURL();
+      // screen.logTestingPlaygroundURL();
 
       expect(screen.getByText(/country/i)).toBeInTheDocument();
       expect(screen.getByText(/organization/i)).toBeInTheDocument();


nmcharlton (commented):
@gwynndp I'm very happy with the org selection refactor.
Are the `CustomTable` changes related?

gwynndp:
> 

The only thing unrelated is that I broke out the Header from CustomTable into it's own component.  I can revert that if we don't want to break components out so we have one component per file.

How do we want to handle cleaning up code to match standards.  I often do it as I go but then it isn't related to the issue directly.  Do we want to create separate issues for cleanup?

nmcharlton:
If the cleaning up is in the same place as the changes, it's fine to include them.
If it's general refactoring, it's best to split it into its own PR, so we can more easily identify where and why things have changed when there is a problem.

src/api/treeTrackerApi.js
```diff
@@ -130,12 +131,15 @@ export default {
    * Capture Match Tool
    */
   fetchCapturesToMatch(currentPage, abortController, filter) {
-    const where = Object.keys(filter)
```
nmcharlton, line 133: Was this not working correctly?

src/api/treeTrackerApi.js
```diff
@@ -130,12 +131,15 @@ export default {
...
-      .map((key) => (filter[key] ? `${key}=${filter[key]}` : ''))
-      .join('&');
+    const where = Object.keys(filter).reduce(
+      (str, key) => (filter[key] ? (str += `&${key}=${filter[key]}`) : str),
+      ''
+    );
+
     const req = `${CAPTURE_MATCH_API}/captures?tree_associated=false&limit=${1}&offset=${
       currentPage - 1
     }&${where}`;
```
nmcharlton, line 141: In the refactored `where` construction, there will always be a `&` at the beginning of the string already.

src/components/EarningsTable/EarningsTable.js
```diff
@@ -123,28 +124,43 @@ function EarningsTable() {
...
   }
 
   async function getEarningsReal(fetchAll = false) {
-    console.warn('fetchAll:', fetchAll);
+    // console.warn('fetchAll:', fetchAll);
+    const filtersToSubmit = { ...filter };
+    // filter out keys we don't want to submit
+    Object.keys(filtersToSubmit).forEach((k) => {
+      if (
+        filtersToSubmit[k] === 'all' ||
```
nmcharlton, line 143: This is a risky way of spotting that all values should be included.
If a user wants to filter by a grower name that starts with or contains "all", we will ignore the value and return all growers.

nmcharlton: I see that it's just been moved from another part of the code, but we should try to resolve it.

nmcharlton (approved):
Looks good to me 👍

src/api/treeTrackerApi.js
gwynndp: I think I ended up with a "&&" at one point, but I'll have to review it again

src/api/treeTrackerApi.js
gwynndp: I'll check this solution again, thanks!

src/components/EarningsTable/EarningsTable.js
gwynndp: I'll make another issue

gwynndp:
@nmcharlton I'm not sure how to get the "npm clean install" step in the CI/CD to pass. Advice?

Siddhant-K-code:
> @nmcharlton I'm not sure how to get the "npm clean install" step in the CI/CD to pass. Advice? It seems to be having trouble with `"react": "*"`, and `"react-dom": "*"` where the Material UI icons are concerned.

there is an inter-dependency issue with webpack versions from different packages.

we can use following command to ignore ^this error:
```yml
npm ci --legacy-peer-deps
```

Siddhant-K-code (approved):

nmcharlton:
@gwynndp I've updated the scripts to use the flag suggested by @Siddhant-K-code. Can you pull the latest changes into your branch and make sure it builds ok?

Siddhant-K-code:
Yeah, it should work now. No need to update branch, just make a empty commit & push it using: 

```
git commit --allow-empty -m "Empty commit"
```

```
git push
```

PR #627, opened 2022-04-04
Add error on failed image load

### LINKED ISSUE #619: Verify tool: Show a useful error on the card when a capture image fails to load
Occasionally, a capture image fails to load for some reason (e.g., timeout, CDN error, image not found).
Currently, it's hard for the user to notice the failure, except for an apparently blank card (the unhelpful image icon is hidden behind the checkbox).
Screenshot 2022-03-28 at 15 42 58

It would be more meaningful to display a fallback "failed to load capture" message on the card, so the user can see something is wrong and either reload the page or report the issue to an admin.

a-maystorov:
## Description

I added an error message when the images fail to load. I have not written any tests for this but it should work everywhere where the `OptimizedImage` component is being used.

**Issue(s) addressed**

- Resolves #619

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**Current behavior**

current

**New behavior**

new

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "treetracker-admin-client",
-  "version": "1.87.8",
+  "version": "1.88.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "treetracker-admin-client",
-      "version": "1.87.8",
+      "version": "1.88.0",
       "dependencies": {
         "@date-io/date-fns": "^1.3.13",
         "@material-ui/core": "^4.9.10",
@@ -9696,9 +9696,9 @@
       }
     },
     "node_modules/async": {
-      "version": "2.6.3",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.3.tgz",
-      "integrity": "sha512-zflvls11DCy+dQWzTW2dzuilv8Z5X/pjfmZOWba6TNIVDm+2UDaJmXSOXlasHKfNBs8oo3M0aT50fDEWfKZjXg==",
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.4.tgz",
+      "integrity": "sha512-mzo5dfJYwAn29PeiJ0zvwTo04zj8HDJj0Mn8TD7sno7q12prdbnasKJHhkm2c1LgrhlJ0teaea8860oxi51mGA==",
       "dependencies": {
         "lodash": "^4.17.14"
       }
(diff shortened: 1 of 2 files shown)


src/components/OptimizedImage.js
```diff
@@ -42,24 +43,52 @@ export default function OptimizedImage(props) {
...
+        <p>
+          Or — <strong>report the issue to an admin.</strong>
+        </p>
+      </div>
+    </Alert>
+  );
+
+  return (
+    <>
+      {!cdnUrl ? (
```
nmcharlton, line 72: This looks like it'll only display the error message when `cdnUrl` is undefined/null.
We also need to fall back when the url is defined but the image fails to load (e.g., error in url).

SirDev97: Hey Nick, I added the native `onError` event handler for images. I set the event `target.src` to `null` in the handler logic, which should trigger the rendering of the fallback message. Tell me, what do you think about this?

nmcharlton: That works nicely.

nmcharlton (changes requested):
The changes look good, but the styling of the message may need tweaking to appear as expected in the Verify tool. It currently renders below where the image should be:
Screenshot 2022-04-18 at 16 02 10

SirDev97:
Hey @nmcharlton the issue was in the styling of the cardContent in the verify tool. I added a fixed height instead of the padding so the image does not get pushed to the bottom. You can see the new behavior below. 

ScreenShot_20220419134423

src/components/Verify.js
```diff
@@ -71,7 +71,8 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.action.selected,
   },
   cardContent: {
-    padding: '87% 0 0 0',
+    // padding: '87% 0 0 0',
+    height: '250px',
```
nmcharlton, line 75: We need to keep the aspect ratio of these cards consistent across different screen size, which is why we were using padding. It may be better to apply `position: absolute` to the `Alert` in `OptimizedImage` to get the desired effect.

SirDev97: @nmcharlton Hey Nick, I made the requested changes and kept the padding. I also tweaked the message styling based on the current location path. I have the images of the different views below, let me know what you think.

capture-matching

verify

growers

src/components/OptimizedImage.js
```diff
@@ -42,24 +44,76 @@ export default function OptimizedImage(props) {
...
-      <img
-        src={cdnUrl || src}
-        alt=".."
-        srcSet={srcSet}
-        sizes={sizes}
-        loading="lazy"
+  const handleImageError = (e) => (e.target.src = null);
+
+  const { pathname } = useLocation();
+  const checkPathName = (path) => pathname === path;
```
nmcharlton, line 50: This is a bit like a circular dependency in disguise.
`OptimizedImage` should not need to know about paths, and using the path to determine how to style the image is likely to cause problems when either the paths change or someone changes or adds image on one of these pages.

Instead, allow the caller to set the style it needs using props. A couple of different options come to mind:
1. Create a set of _modes_ or _roles_ that use pre-defined styles and allow the caller to specify which to use as a prop.
2. Pass these usage-specific style override attributes as props to `OptimizedImage` (e.g., `imgStyle`, `alertStyle`, `alertTitleStyle`).

SirDev97: I understand your concerns. I'll see what I can do and maybe encapsulate the logic in a new ImageError component.

PR #635, opened 2022-04-06
refactor: remove icon log warning from reportingCard

### LINKED ISSUE #622: Remove unnecessary warning logging from Reporting cards
Some development logging has been left over in the dashboard reporting cards and can be removed:

Screenshot 2022-04-03 at 14 23 00

a-maystorov:
**Issue(s) addressed**

- Resolves #622

- [ ] Enhancement
- [ ] Bug fix
- [x] Refactor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.87.7](https://github.com/Greenstand/treetracker-admin-client/compare/v1.87.6...v1.87.7) (2022-04-07)
+
+
+### Bug Fixes
+
+* timezone problem ([e618f75](https://github.com/Greenstand/treetracker-admin-client/commit/e618f75921012cbfaf67e1ca5799daad0922e4b1))
+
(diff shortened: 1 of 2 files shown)


src/components/reportingCards/ReportingCard.js
```diff
@@ -145,7 +145,7 @@ function GrowerReportingCard(props) {
     setOpen(true);
   }
 
-  log.warn('icon', icon);
+  // log.warn('icon', icon);
```
TANguyen1893, line 148: What do you think about removing it completely, rather than commenting it out? I would prefer to have it removed so we don't have useless comments in the code.

SirDev97: Sure thing. I didn't know if we would need it for future tests or something; that's why I didn't delete it, but yeah makes sense to remove it altogether.

TANguyen1893 (approved):

PR #652, opened 2022-04-24
Verify, Growers, CaptureMatching tools: Showing an error message when images fail to load

### LINKED ISSUE #619: Verify tool: Show a useful error on the card when a capture image fails to load

a-maystorov:
## Description

I added a new `ImageErrorAlert` component with usage-specific props to style it based on where the `OptimizedImage` component is being used. I also handled the native `onError` event handler for the img in `OptimizedImage.` I set the event target src to null, which should trigger the rendering of the `ImageErrorAlert` component.

**Issue(s) addressed**

- Resolves #619

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

**New behavior**

verify

capture

growers

### DIFF
--- a/src/components/CaptureMatching/CandidateImages.js
+++ b/src/components/CaptureMatching/CandidateImages.js
@@ -223,7 +223,7 @@ function CandidateImages({ capture, candidateImgData, sameTreeHandler }) {
                           className={classes.candidateCaptureContainer}
                         >
                           <OptimizedImage
-                            src={'candidateCapture.image_url'}
+                            src={candidateCapture.image_url}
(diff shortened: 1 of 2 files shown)


nmcharlton (approved):
This is excellent, thanks @SirDev97!

PR #655, opened 2022-05-03
feat(captures): add token id column in list view

### LINKED ISSUE #653: "Captures" module needs token_id  in list view
For organisations to better manage their data themselves it would be super helpful if the token_id is included in the list view since it is already in the filters. 
Currently you can filter by tokenized or not but exporting you only get the capture ids which does not allow to create a token request query since that is always based on the token_id.

Am giving this a priority status since it should be easy to fix and logically token_ids enables actual wallet transfers to be generated from the captures tool. In the future we need to include wallet name and wallet id as well.

Mloweedgar:
## Description

add token id column in captures list view 

**Issue(s) addressed**

- Resolves #653 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/components/Captures/CaptureTable.js
+++ b/src/components/Captures/CaptureTable.js
@@ -1,8 +1,8 @@
 import React, { useEffect, useState, useContext } from 'react';
 import {
+  Button,
   Grid,
   Table,
-  Button,
   TableHead,
   TableBody,
   TableRow,
@@ -10,6 +10,7 @@ import {
   TablePagination,
   TableSortLabel,
   Typography,
+  Tooltip,
 } from '@material-ui/core';
 import { GetApp } from '@material-ui/icons';
 import { getDateTimeStringLocale } from '../../common/locale';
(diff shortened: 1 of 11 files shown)


src/components/Captures/CaptureTable.js
```diff
@@ -59,6 +59,13 @@ const columns = [
...
     renderer: (val) => val,
   },
+
+  {
+    attr: 'tokenId',
+    label: 'Token Id',
+    renderer: (val) => val,
+  },
+
   {
```
nmcharlton, line 69: Let's remove the Token Status column as it's obvious from the Token ID.

src/components/Captures/CaptureTable.js
```diff
@@ -59,6 +59,13 @@ const columns = [
     noSort: true,
     renderer: (val) => val,
   },
+
+  {
+    attr: 'tokenId',
+    label: 'Token Id',
```
nmcharlton, line 65: Should be `'Token ID'`

nmcharlton (changes requested):

nmcharlton:
Can we fix the failing tests?

nmcharlton (approved):

PR #657, opened 2022-05-11
fix: remove email and phone number from edit grower dialog

### LINKED ISSUE #239: Bug: When planter profiles get edited through the admin panel it breaks the field schema ingestion
I have just come across a case where an admin panel user changed the email address of a planter which resulted in the App data to get stuck in the "bulk_tree_upload" since there was no match of the identifier. Another case where i had edited multiple planter entries to match the telephone number resulted in the same issue.

rishad190:
## Description

**Issue(s) addressed**

- Resolves #239 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

**After fix**

### DIFF
--- a/src/components/EditGrower.js
+++ b/src/components/EditGrower.js
@@ -124,20 +124,6 @@ const EditGrower = (props) => {
         label: 'Last Name',
       },
     ],
-    [
-      {
-        attr: 'email',
-        label: 'Email Address',
-        type: 'email',
-      },
-    ],
-    [
-      {
-        attr: 'phone',
-        label: 'Phone Number',
-        type: 'tel',
-      },
-    ],
   ];
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/EditGrower.js
```diff
@@ -158,21 +158,25 @@ const EditGrower = (props) => {
...
-                  InputLabelProps={{
-                    shrink: true,
-                  }}
-                  onChange={(e) => {
-                    handleChange(input.attr, e.target.value);
-                  }}
-                  value={getValue(input.attr)}
-                />
+                <>
+                  {!(input.type === 'email') && !(input.type === 'tel') && (
```
nmcharlton, line 162: Rather than conditionally omitting inputs, it would be best to remove these items from the `inputs` array altogether. These values are not used anywhere else.

nmcharlton (changes requested):
Thanks for picking this up, @rishad190!

src/components/EditGrower.js
rishad190: @nmcharlton done . Look into this solution.

PR #658, opened 2022-05-14
fix: forget route when a user logs out explicit

### LINKED ISSUE #624: Forget route when a user logs out explicitly
Some of our users have more than one Admin Panel account (e.g., to support multiple organizations).
When a user logs out of one account (from the Account page) and logs in with another, they are redirected to the last page they were on (Account) which seems like odd behaviour.

When they log out explicitly by clicking _Log Out_, then log back in, we should redirect them to the home page.

Note that the redirection to the last page is desired behaviour on an _implicit_ logout (e.g., token expiry) to minimise disruption. This should be retained.

rishad190:
## Description

**Issue(s) addressed**

- Resolves #624 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/Login.js
+++ b/src/components/Login.js
@@ -66,8 +66,8 @@ const Login = (props) => {
 
   const history = useHistory();
   const location = useLocation();
-  const { from } = { from: { pathname: '/' } };
-  console.log(location);
+  const { from } = location.state || { from: { pathname: '/' } };
+
   useEffect(() => {
     return () => {
       setLoading(false);
--- a/src/components/Account.js
+++ b/src/components/Account.js
@@ -19,6 +19,7 @@ import { AppContext } from '../context/AppContext';
 import axios from 'axios';
 import { getDateTimeStringLocale } from '../common/locale';
 import { documentTitle } from '../common/variables';
+import { useHistory } from 'react-router-dom';
 
 const style = (theme) => ({
   accountContainer: {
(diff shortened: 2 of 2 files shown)


src/components/Login.js
```diff
@@ -66,8 +66,8 @@ const Login = (props) => {
 
   const history = useHistory();
   const location = useLocation();
-  const { from } = location.state || { from: { pathname: '/' } };
-
+  const { from } = { from: { pathname: '/' } };
+  console.log(location);
```
nmcharlton, line 70: You can remove `console.log()` when development is complete so we don't pollute the console in production. If you need to keep logging in production, use `loglevel` instead.

src/components/Login.js
```diff
@@ -66,8 +66,8 @@ const Login = (props) => {
 
   const history = useHistory();
   const location = useLocation();
-  const { from } = location.state || { from: { pathname: '/' } };
```
nmcharlton, line 69: I can't test this at the moment, but it looks like this will cause us to _always_ forget the path. Instead, we should only forget it when a user _explicitly_ logs out via the Account page.

If they are redirected to the login page (e.g., because their session expired or they weren't logged in to begin with), the path they entered should be remembered.

nmcharlton (changes requested):

rishad190:
@nmcharlton now it's working fine. You can check now.

nmcharlton (approved):
Nice work 😄

PR #685, opened 2022-06-09
feat: outline removable capture tag chips

### LINKED ISSUE #675: Capture Details: Allow user to remove capture tags
If a user has the `APPROVE_TREE` policy, they should be able to remove capture tags, one by one, from captures via the Capture Details dialog:

1. Capture tag chips (the last ones listed under _Other_) are marked deletable for privileged users (X appears to the right of the label)
2. Clicking the X prompts _Are you sure you wish to remove {name} from this capture?_
3. User clicks _Remove Tag_ to confirm (_Cancel_ option also available).
4. Tag is removed from the capture by calling `DELETE /api/tree_tags/:id`.
5. Tag disappears from the list of chips.

Screenshot 2022-06-02 at 14 20 55

a-marquez:
## Description

* Display deletion control on removable capture tags when `appContext.user` has `APPROVE_TREE` policy.
* Separate non-removable tags and capture tags rendering. 

**Issue(s) addressed**

- Resolves #675

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

* Tags are displayed as plain chips

**What is the new behavior?**
* Capture tag chips have a remove icon when appropriate
* Removal requires confirmation

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -314,9 +314,8 @@ function CaptureDetailDialog(props) {
                     label={tag.tagName}
                     className={classes.chip}
                     onDelete={
-                      hasApproveTreePermission !== undefined // TODO: delete, for testing purpose only
-                        ? // onDelete={(hasApproveTreePermission
-                          () => {
+                      hasApproveTreePermission === true
+                        ? () => {
                             setCaptureTagDeletionTarget({
                               capture: cdContext.capture,
                               tag,
--- a/package-lock.json
+++ b/package-lock.json
@@ -36958,7 +36958,7 @@
       "version": "2.25.0",
       "resolved": "https://registry.npmjs.org/webpack-hot-middleware/-/webpack-hot-middleware-2.25.0.tgz",
       "integrity": "sha512-xs5dPOrGPCzuRXNi8F6rwhawWvQQkeli5Ro48PRuQh8pYPCPmNnltP9itiUPT4xI8oW+y0m59lyyeQk54s5VgA==",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "ansi-html": "0.0.7",
         "html-entities": "^1.2.0",
@@ -36970,7 +36970,7 @@
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
       "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-      "dev": true,
+      "devOptional": true,
       "engines": {
         "node": ">=0.10.0"
       }
@@ -36979,7 +36979,7 @@
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
       "integrity": "sha1-ajhfuIU9lS1f8F0Oiq+UJ43GPc8=",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "ansi-regex": "^2.0.0"
       },
@@ -40736,7 +40736,8 @@
     "@material-ui/types": {
       "version": "5.1.0",
       "resolved": "https://registry.npmjs.org/@material-ui/types/-/types-5.1.0.tgz",
-      "integrity": "sha512-7cqRjrY50b8QzRSYyhSpx4WRw2YuO0KKIGQEVk5J8uoz2BanawykgZGoWEqKm7pVIbzFDN0SpPcVV4IhOFkl8A=="
+      "integrity": "sha512-7cqRjrY50b8QzRSYyhSpx4WRw2YuO0KKIGQEVk5J8uoz2BanawykgZGoWEqKm7pVIbzFDN0SpPcVV4IhOFkl8A==",
+      "requires": {}
     },
     "@material-ui/utils": {
       "version": "4.11.2",
@@ -40832,7 +40833,8 @@
     "@rematch/core": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/@rematch/core/-/core-2.0.1.tgz",
-      "integrity": "sha512-EyvgmXzXZnDCIIsmBmOtwzaTQzH+rAlrjcPTiV8Hd/TbFeH5VpCXI/B4myB5aH4P+yQafRkm846HNHMmhkq7kQ=="
+      "integrity": "sha512-EyvgmXzXZnDCIIsmBmOtwzaTQzH+rAlrjcPTiV8Hd/TbFeH5VpCXI/B4myB5aH4P+yQafRkm846HNHMmhkq7kQ==",
+      "requires": {}
     },
     "@rollup/plugin-node-resolve": {
       "version": "7.1.3",
@@ -44951,7 +44953,8 @@
     "acorn-jsx": {
       "version": "5.3.1",
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-5.3.1.tgz",
-      "integrity": "sha512-K0Ptm/47OKfQRpNQ2J/oIN/3QYiK6FwW+eJbILhsdxh2WTLdl+30o8aGdTbm5JbffpFFAg/g+zi1E+jvJha5ng=="
+      "integrity": "sha512-K0Ptm/47OKfQRpNQ2J/oIN/3QYiK6FwW+eJbILhsdxh2WTLdl+30o8aGdTbm5JbffpFFAg/g+zi1E+jvJha5ng==",
+      "requires": {}
     },
     "acorn-node": {
       "version": "1.8.2",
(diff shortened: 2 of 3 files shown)


a-marquez:
Screenshot 2022-06-09 061229

I'm at the moment substituting actual `captureTag` data with a mock property, I'm uncertain as to which actual capture to test against. I thought to use the tag filter with `photoless` but oddly that tag itself doesn't display in the `CaptureDetailDialog`.

src/components/CaptureDetailDialog.js
```diff
@@ -265,14 +297,70 @@ function CaptureDetailDialog(props) {
...
+                    alignItems: 'center',
+                    padding: '0px',
+                    marginTop: '1em',
+                  }}
+                >
+                  <Typography>
+                    Remove tag <b>{`"${captureTagDeletionTarget}"`}</b> ?
+                  </Typography>
+
+                  <ButtonGroup>
```
nmcharlton, line 340: I'd suggest using two separate buttons with a gap rather than a `ButtonGroup`. This looks more like a toggle switch, and no gap makes mis-clicks more likely.

src/components/CaptureDetailDialog.js
```diff
@@ -115,23 +135,35 @@ function CaptureDetailDialog(props) {
...
 
   function Tags(props) {
     const { capture, species, captureTags } = props;
-    const allTags = [
+    const otherTags = [
       capture.morphology,
       capture.age,
       capture.captureApprovalTag,
       capture.rejectionReason,
-      ...captureTags.map((t) => t.tagName),
```
nmcharlton, line 132: This should contain deletable tags, but only gets populated after the dialog is opened as the tag names are queried from the server.

nmcharlton:
@a-marquez This is looking really good.
You can test against any tag you can find in the data. `photoless` is a good candidate, but it's odd that it doesn't appear in the CaptureDetailsDialog for you. It should appear a second or so after opening the dialog:
Screenshot 2022-06-11 at 21 43 08

src/components/CaptureDetailDialog.js
a-marquez: Understood, I believe that the `otherTags` collection is currently only string values that get rendered. I'm separating the capture tags to render with respective api IDs in preparation for the deletion call on removal confirmation.

a-marquez:
@nmcharlton Thank you! Regarding the API side of this, in ticket #675 you specify the call being `DELETE /api/tree_tags/:id` but shouldn't it be `/captures/:capture_id/tags/:tag_id`? I'm concerned `DELETE /api/tree_tags/:id` would delete all instances of a tag instead of the individual capture's tag.

I also don't actually see deletion implemented, I'm happy to take it on. I'm looking at at [`routes.js`](https://github.com/Greenstand/treetracker-api/blob/77961f6912968321c5f79b017c5a7493e57c584f/server/routes.js#L104) and  [`CaptureTagService`](https://github.com/Greenstand/treetracker-api/blob/main/server/services/CaptureTagService.js).

nmcharlton:
Hey @a-marquez. Do you need any support to move this PR forward?

a-marquez:
Hi @nmcharlton, following our last call you assured me that `DELETE /api/tree_tags/:id` is the correct call (referencing ) and that it will delete the relation between the capture and the tag, not the tag itself. I just found it a little odd that both captures with the `photoless` tag have the same `tag.id`. Currently, I expect if I make the specified call that the tag will be deleted from both captures (see below).

Screenshot from 2022-07-10 10-10-19

I will update once I've connected to the DB so that I can verify & revert if needed. Please advise if I should do anything differently.

nmcharlton:
Hi @a-marquez. To try to clarify:

There are three entities, each with its own `id`:
- `tree` (referred to as "capture" in the client)
- `tag`, e.g. _photoless_
- `tree_tag`, the relationship between a `tree` and a `tag`

In the example you show, the `tag` has `id` 23, but each of the two `tree_tag` relationships has a unique `id`.
That's the value to pass to `DELETE /api/tree_tag/:id`

If `tree_tag.id` isn't available, you can query `GET /api/tree_tags?filter[where][tree_id]=878654&filter[where][tag_id]=23` 

I hope that makes sense.

a-marquez:
@nmcharlton yes, thank you, I was taking it for granted that `tag.id` was meant to be the unique id. I'll continue as you described.

nmcharlton:
Sorry for the lack of clarity in the code, that's my bad.

a-marquez:
@nmcharlton Not a problem! I know naming can get messy. I've pushed up the latest code and have a few notes. 

* Attempting to get a single `tree_tag.id` using `treeTrackerApi.getCaptureTags()` was responding with a syntax error after requesting. I am figuring the query clause builders in [`treeTrackerApi.getCaptureTags`](https://github.com/a-marquez/treetracker-admin-client/blob/ccad8a56b85193d521774cbd0ef0d9cd9ffef2e1/src/api/treeTrackerApi.js#L448) can't currently handle a single captureTagId/tagId query. I have added a function [`getSingleCaptureTag`](https://github.com/a-marquez/treetracker-admin-client/blob/ccad8a56b85193d521774cbd0ef0d9cd9ffef2e1/src/api/treeTrackerApi.js#L429) for now.

* I believe the request is firing correctly but I'm receiving a `401 Unauthorized` error. I believe this may be because the user I'm logged in with doesn't have the approve tree permission? I'd appreciate some direction on this, I have written in the policy permission check but for development purposes have just been bypassing it so far. Please advise on how to continue with this.

Screenshot from 2022-07-12 19-21-49

nmcharlton:
> * Attempting to get a single tree_tag.id using treeTrackerApi.getCaptureTags() was responding with a syntax error after requesting. I am figuring the query clause builders in [treeTrackerApi.getCaptureTags](https://github.com/a-marquez/treetracker-admin-client/blob/ccad8a56b85193d521774cbd0ef0d9cd9ffef2e1/src/api/treeTrackerApi.js#L448) can't currently handle a single captureTagId/tagId query. I have added a function [getSingleCaptureTag](https://github.com/a-marquez/treetracker-admin-client/blob/ccad8a56b85193d521774cbd0ef0d9cd9ffef2e1/src/api/treeTrackerApi.js#L429) for now.

That makes sense.

> * I believe the request is firing correctly but I'm receiving a 401 Unauthorized error. I believe this may be because the user I'm logged in with doesn't have the approve tree permission? I'd appreciate some direction on this, I have written in the policy permission check but for development purposes have just been bypassing it so far. Please advise on how to continue with this.

You're missing the `s` off the end of `tree_tags` in the URL.
We don't actually apply policy restrictions to the `tree_tags` route, so it should work once the URL is fixed.

a-marquez:
@nmcharlton Embarrassing! Yes the call is working properly now, apologies for that, I'm now just going to make sure it correctly refreshes the capture detail context and the underlying table before proceeding.

src/components/CaptureDetailDialog.js
```diff
@@ -265,14 +299,87 @@ function CaptureDetailDialog(props) {
...
+                  <Chip key={tag} label={tag} className={classes.chip} />
+                ))}
+
+                {captureTags.map((tag) => (
+                  <Chip
+                    key={tag.tagName}
+                    label={tag.tagName}
+                    className={classes.chip}
+                    onDelete={
+                      hasApproveTreePermission !== undefined // TODO: delete, for testing purpose only
```
nmcharlton, line 317: As per the comment, should this be deleted?

a-marquez: Yes, thank you, I have fixed it to correctly check for the APPROVE_TREE permission.

nmcharlton (approved):

nmcharlton (changes requested):
@a-marquez There's a slight issue with captures that have more than one tag. When the capture details reload after removing one, the remaining tags are missing.

They appear when the capture details drawer is reopened:

https://user-images.githubusercontent.com/5558838/180664033-c52f89ce-827f-4d13-8b22-43bcb5e24a15.mov

nmcharlton:
@a-marquez Have you been able to make any progress on this?

mayankbucha:
@nmcharlton how can I add multiple capture tags on a capture?

nmcharlton:
> @nmcharlton how can I add multiple capture tags on a capture?

In the Verify tool, select a capture and input two (or more) items in the Additional tags field and press Submit.
Screenshot 2022-09-26 at 21 50 56

mayankbucha:
@nmcharlton the issue is in the function `handleCaptureTagDeletion`.

```
async function handleCaptureTagDeletion({ capture, tag }) {
    try {
      await cdContext.deleteCaptureTag({
        captureId: capture?.id,
        tagId: tag.id,
      });
      if (props.onCaptureTagDelete !== undefined) {
        props.onCaptureTagDelete();
      }
    } catch (error) {
      console.log(error);
    }
    setCaptureTagDeletionTarget(undefined);
  }
```

Here, the `props.onCaptureTagDelete` is getting undefined due to which the tags cannot be loaded from the parent component. I cannot figure out why this function is getting undefined.
Should I try any other approach?

src/components/CaptureDetailDialog.js
```diff
@@ -115,21 +134,36 @@ function CaptureDetailDialog(props) {
...
 
+  async function handleCaptureTagDeletion({ capture, tag }) {
+    try {
+      await cdContext.deleteCaptureTag({
+        captureId: capture?.id,
+        tagId: tag.id,
+      });
+      if (props.onCaptureTagDelete !== undefined) {
+        props.onCaptureTagDelete();
+      }
```
mayankbucha, line 145: onCaptureTagDelete is showing undefined.

src/components/CaptureDetailDialog.js
```diff
@@ -82,11 +94,18 @@ function CaptureDetailDialog(props) {
   // console.log('render: capture detail dialog');
   const { open, capture } = props;
   const cdContext = useContext(CaptureDetailContext);
+  const appContext = useContext(AppContext);
+  const hasApproveTreePermission = hasPermission(appContext.user, [
+    POLICIES.APPROVE_TREE,
```
nmcharlton, line 99: Need to include `POLICIES.SUPER_PERMISSION`

mayankbucha: I Have added this permission but the `onCaptureTagDelete` function is still undefined on the props.

src/components/CaptureDetailDialog.js
nmcharlton: It looks like `onCaptureTagDelete` is an optional callback to the calling component to make any updates it needs to.
In the case of _Verify_ there are no component updates necessary, so it makes sense that this is `undefined`.

The problem appears to be that `CaptureDetailContext` is not updating properly when a tag is deleted. It should refresh the capture detail when deletion is complete.

mayankbucha: Ok, I'll check it.

nmcharlton: Hey @mayankbucha. Did you ever manage to make any progress on completing this PR?

nmcharlton:
Looks like we're struggling to finish this one off – I'll see if I can move it forward when I get a chance.

nmcharlton:
Replaced by #986

PR #748, opened 2022-07-19
Add org filter working for growers, captures, verify, capture match

gwynndp:
## Description

 - Add org filter working for growers, captures, verify, capture match
 - Depends on Query APi: https://github.com/Greenstand/treetracker-query-api/pull/132

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/context/AppContext.js
+++ b/src/context/AppContext.js
@@ -37,6 +37,7 @@ import { session, hasPermission, POLICIES } from '../models/auth';
 import api from '../api/treeTrackerApi';
 import stakeholder_api from '../api/stakeholders';
 import RegionsView from 'views/RegionsView';
+import log from 'loglevel';
 
 // no initial context here because we want login values to be 'undefined' until they are confirmed
 export const AppContext = createContext({});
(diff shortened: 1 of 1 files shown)


src/context/AppContext.js
```diff
@@ -305,15 +305,18 @@ export const AppProvider = (props) => {
...
     const { stakeholders } = await stakeholder_api.getStakeholders();
-    const results = stakeholders.map((s) => {
-      return {
-        id: s.id,
-        stakeholder_uuid: s.id,
-        name: s.org_name || s.first_name + ' ' + s.last_name,
-        type: s.type,
-      };
+    const results = [];
+    stakeholders.forEach((s) => {
```
nmcharlton, line 309: This isn't populating `results`

nmcharlton (changes requested):

src/context/AppContext.js
nmcharlton: @gwynndp Can you double check this method when you get a chance?
It appears to duplicate the orgs in the org list (assuming Stakeholders API has the same data).

nmcharlton (changes requested):
We need to add `REACT_APP_STAKEHOLDER_API_ROOT` to `.env.test`, `.env.staging` and `.env.production`

src/context/AppContext.js
gwynndp: I adjusted it to only show the stakeholders that the user is allowed to see.    It may mean that temporarily we won't see Person type stakeholders in the list or any legacy orgs that don't have a corresponding stakeholder  with the same uuid.  (Right now there is only one, a duplicate "test2" org)

nmcharlton (approved):

PR #762, opened 2022-07-25
chore: Truncate lat/lon and image URL #761

### LINKED ISSUE #761: Captures: Truncate lat/lon and image URL
There's no need to display the whole image URL and co-ordinates for captures in the Captures table.
We should truncate the lat and lon (six decimal places is enough) and replace the long image URL with a link (similar to what we have in the Capture Details dialog.

Note that the raw values should still be included in the CSV export.

Screenshot 2022-07-25 at 21 13 52

iib0011:
Resolves #761


nmcharlton (approved):
Nice work @iib0011!

Issue #773, opened 2022-08-04
Tidy reporting card titles


nmcharlton:
Wording, capitalisation and pluralisation is inconsistent in reporting cards.

The following changes should be made:
* "Grower" -> "Growers"
* "total" -> "Total"
* "Average captures per grower per organization" -> "Average Captures per Grower" (per organization is implied)
* On Species card, "Total" -> "Count"
<img width="1552" alt="Screenshot 2022-08-04 at 20 57 22" src="https://user-images.githubusercontent.com/5558838/182941722-d468a4b3-fc6f-4e41-8244-df89c3f6657e.png">

muhon9:
Will fix it

nmcharlton:
> Will fix it
> 
> 

Sorry @muhon9, @jwardoin  managed to get the fix in quickly on this one.

PR #774, opened 2022-08-05
fix: tidied reporting card titles

jwardoin:
In reference to [Issue #773](https://github.com/Greenstand/treetracker-admin-client/issues/773)


nmcharlton (approved):
Thanks @jwardoin

PR #775, opened 2022-08-05
fix: api request for Capture Detail from Capture Match vs. other tools

gwynndp:
## Description

**Issue(s) addressed**

- Resolves #issue_number

- [ ] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -88,12 +88,11 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 function CaptureDetailDialog(props) {
-  log.debug('render: capture detail dialog');
-  const { open, capture, onClose } = props;
+  const { open, captureId, onClose } = props;
   const cdContext = useContext(CaptureDetailContext);
   const [snackbarOpen, setSnackbarOpen] = useState(false);
   const [snackbarLabel, setSnackbarLabel] = useState('');
-  const [renderCapture, setRenderCapture] = useState(capture);
+  const [renderCapture, setRenderCapture] = useState(null);
   const [screenWidth, setScreenWidth] = useState(window.innerWidth);
   const [screenHeight, setScreenHeight] = useState(window.innerHeight);
   const resizeWindow = useCallback(() => {
@@ -103,12 +102,11 @@ function CaptureDetailDialog(props) {
   const classes = useStyles();
 
   useEffect(() => {
-    // prevent request if capture is empty or undefined because it will fail without a valid id
-    // but if we have capture info, do we need to make the request?
-    if (capture?.id) {
-      cdContext.getCaptureDetail(capture);
+    // prevent request because it will fail without a valid id
+    if (captureId) {
+      cdContext.getCaptureDetail(captureId);
     }
-  }, [capture]);
+  }, [captureId]);
 
   useEffect(() => {
     window.addEventListener('resize', resizeWindow);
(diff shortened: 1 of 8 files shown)


src/components/CaptureDetailDialog.js
```diff
@@ -114,10 +121,35 @@ function CaptureDetailDialog(props) {
...
+        captureApprovalTag: current.captureApprovalTag || null,
+        morphology: current.morphology || null,
+        note: 'My Note',
+        rejectionReason: current.rejectionReason || null,
+        species_id: current.species_id || current.speciesId,
+        token_id: current.token_id || current.tokenId,
+        created_at: current.created_at || current.timeCreated,
+        updated_at: current.updated_at || current.timeUpdated,
+      });
     }
```
gwynndp, line 153: Manage different keys from old and new apis by consolidating into a new object.  The consolidation matches the new api.

src/context/CaptureDetailContext.js
```diff
@@ -36,15 +41,45 @@ export function CaptureDetailProvider(props) {
...
+    log.debug('getCaptureDetail:', {
+      reference_id: capture.reference_id,
+      id: capture.id,
+    });
+
+    // getOrganization gets the wrong org when this is in the global context
+    const BASE_URL = {
+      LEGACY: `${API_ROOT}/api/${getOrganization()}trees/`,
+      CAPTURE_MATCH: `${TREETRACKER_API}/captures?reference_id=`,
+    };
```
gwynndp, line 54: This can be used to determine which microservice to make the request to   in the future, but we'll have to pass a second argument to indicate which tool initiated the request.

Seem like a solution?

src/context/CaptureDetailContext.js
```diff
@@ -36,15 +41,45 @@ export function CaptureDetailProvider(props) {
...
+        })
+          .then(handleResponse)
+          .then((data) => {
+            if (data.captures) {
+              // new treetracker-api
+              setState({ ...state, capture: data.captures[0] });
+            } else {
+              // legacy api
+              setState({ ...state, data });
+            }
```
gwynndp, line 78: The legacy api doesn't return the capture nested in an object so we need this to handle the difference.

src/components/CaptureDetailDialog.js
```diff
@@ -100,7 +103,11 @@ function CaptureDetailDialog(props) {
   const classes = useStyles();
 
   useEffect(() => {
-    cdContext.getCaptureDetail(capture?.id);
+    // prevent request if capture is empty or undefined because it will fail without a valid id
+    // but if we have capture info, do we need to make the request?
+    if (capture?.id) {
+      cdContext.getCaptureDetail(capture);
+    }
   }, [capture]);
```
gwynndp, line 111: I think Dialogs in Material UI might be designed to receive props.  If we pass props then we are already getting all the info we need.  If we don't pass any props we wouldn't have the id to make the request.  We could pass the id for the request through a route, possibly, but I think we'd have to add in react router or something else to support the detail view.

nmcharlton: As discussed, it's best if the Capture Detail dialog can be populated when passed just a reference ID, raw capture ID or processed capture ID. The dialog (or corresponding context) can then build up a complete picture of the capture, its status and associated wallet registration, device info etc.

There's a case to move this aggregation to a Capture Query API, as a lot of the data has been requested in the Captures list as well.

nmcharlton (approved):
Thanks @gwynndp, I really like this solution.
Sorry for the delay in reviewing.

src/context/CaptureDetailContext.js
nmcharlton: Looks good to me.

nmcharlton:
@gwynndp Are you happy for this to be merged? We'll need it before we can do any more releases.

gwynndp:
Yes.  The only hold-up is that I'm trying to update the tests and having
difficulties, but I can make a separate PR once I figure it out.

Gwynn

PR #776, opened 2022-08-08
fix: filtering by multiple Grower ids gives incorrect results

### LINKED ISSUE #766: Captures: Filtering by multiple Grower IDs gives incorrect results
## Steps to reproduce
1. Open Captures tool
2. Enter two Grower IDs (select from the Captures list) separated by a comma in the _Grower ID_ filter and press _Apply_
3. Captures from other growers are included in the results.

Screenshot 2022-07-30 at 21 12 06

rutvij2811:
## Description

_[Filtering by multiple Grower IDs gives incorrect results]_
### Steps to reproduce
Open Captures tool
Enter two Grower IDs (select from the Captures list) separated by a comma in the Grower ID filter and press Apply
Captures from other growers are included in the results.
181994776-ded46afe-b7b8-4b69-b16a-5502767c24f7

**Issue(s) addressed**

- Resolves #766

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Filtering by multiple Grower IDs gives incorrect results

**What is the new behavior?**
Filtering by multiple Grower IDs gives correct results

## Other useful information

Had to make changes to the Filter.js from models which take into consideration adding "AND" and "OR" clauses to the "where" query based on the number of IDS passed as well as the number of "verification status" options selected.
The previous version clubbed everything under a single "OR" clause due to which we received incorrect results.

Results after the fix:
greenstandFixPR

### DIFF
--- a/src/models/Filter.test.js
+++ b/src/models/Filter.test.js
@@ -65,7 +65,13 @@ describe('Filter, with initial values about this filter object', () => {
 
   it('getWhereObj() should match: planterId=1', () => {
     expect(filter.getWhereObj()).toEqual(
-      expect.objectContaining({ planterId: '1' })
+      expect.objectContaining({
+        and: expect.arrayContaining([
+          {
+            or: [{ planterId: '1' }],
+          },
+        ]),
+      })
     );
   });
 


src/models/Filter.js
```diff
@@ -138,6 +138,13 @@ export default class Filter {
         if (!orCondition) {
           orCondition = true;
           where = [];
+          // Incase only 1 verify status selected, then the above statement overwrites the previous "where" status.
```
nmcharlton, line 141: This method is starting to become quite difficult to follow and easy to break.
LoopBack supports passing a single item to an `or` statement. Perhaps we should _always_ assume we will use `or` for `verifyStatus` and `planterId`, regardless of the number of values.

The database should optimise the query anyway, so I don't think it will have a negative impact on performance.

rutvij2811: Got it. With that assumption, there is no need for an "orCondition" flag as it Is always going to be true.
I will work on that. Will the new commit automatically be added to this PR?

nmcharlton: Yes, just push additional commits and the PR will be updated.

rutvij2811: @nmcharlton , I have pushed the new commit. Requesting your review for the same.

src/models/Filter.js
```diff
@@ -110,15 +110,16 @@ export default class Filter {
       where.verifyStatus = this.verifyStatus;
     }
 
-    let orCondition = false;
     const { verifyStatus, ...restFilter } = where;
 
     if (verifyStatus) {
       if (verifyStatus.length === 1) {
```
nmcharlton, line 116: This clause can be removed too.
I don't like the reuse of `where` here – we'd be better off with something clearer. (Sorry, I know it's not your code originally. This would just be a good time to tidy it up.)

In fact, we don't need to do process `verifyStatus` at all, and the `planterIds` can be simplified to:
```
const planterIds = this.planterId.split(',').map((item) => ({
  planterId: item.trim());
});
```
Then we can just return this:
```
return { ...restFilter, or: verifyStatus, or: planterIds };
```
We will just need to check how LoopBack copes with empty `or` arrays.

nmcharlton (changes requested):

src/models/Filter.js
rutvij2811: I got your point.
Let me work on it. I'll let you know once I commit the new changes.

rutvij2811: @nmcharlton ,
I have done the refactoring.

nmcharlton (approved):
@rutvij2811 I hope you don't mind, I did a little extra refactoring to simplify the logic further.
This should make it easily extensible for other fields.

PR #780, opened 2022-08-12
Add gender report card for Freetown

### LINKED ISSUE #703: GENDER. One of the most important requirement by Freetown is to show gender percentage in planters
In order to show this we need to have 3 genders. male/female/neutral. This is a critical requirement!

gwynndp:
## Description

Add report card with gender statistics

https://monosnap.com/file/8OlMO8xDggeFtbmmPhRdzpUpJcxiyq

**Issue(s) addressed**

- Resolves #703
- Depends on PR for the reporting api (https://github.com/Greenstand/treetracker-reporting/pull/50)

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/reportingCards/ReportingCard.js
+++ b/src/components/reportingCards/ReportingCard.js
@@ -178,7 +178,6 @@ function ReportingCard(props) {
                 <Typography className={classes.name}>{item.name}</Typography>
                 <Typography className={classes.number}>
                   {new Intl.NumberFormat().format(item.number)}
-                  {item.percentage && `  (${item.percentage}%)`}
                 </Typography>
               </Box>
             ))
@@ -243,7 +242,6 @@ function ReportingCard(props) {
                     </Typography>
                     <Typography className={classes.number}>
                       {new Intl.NumberFormat().format(item.number)}
-                      {item.percentage && `  (${item.percentage}%)`}
                     </Typography>
                   </Box>
                 ))}
--- a/src/components/reportingCards/ReportingCard8.js
+++ b/src/components/reportingCards/ReportingCard8.js
@@ -38,7 +38,7 @@ export default function component(props) {
   return (
     <ReportingCard
       text={{
-        title: 'Gender Percentage',
+        title: 'Grower Gender Counts',
         text1: 'Total Growers',
       }}
       icon={PeopleIcon}


src/components/reportingCards/ReportingCard8.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const formatData = (data) => {
+    const result = data.top.map((item) => {
+      return {
+        name:
+          item.name && item.name.match(/f/i)
+            ? 'Female'
+            : item.name
+            ? 'Male'
+            : 'Unknown',
```
nmcharlton, line 27: In the database/API we should support _Female_, _Male_ and _Non-binary_

.env.development
```diff
@@ -14,3 +14,4 @@ REACT_APP_ENABLE_MESSAGING=true
 REACT_APP_ENABLE_REGIONS=true
 REACT_APP_ENABLE_TOOLTIPS=true
 REACT_APP_ENABLE_STAKEHOLDERS=true
+REACT_APP_ENABLE_GENDER_REPORTING_CARD=true
```
nmcharlton, line 17: The API is already deployed to production. I don't think we need a feature flag for this.

src/components/reportingCards/ReportingCard8.js
```diff
@@ -0,0 +1,46 @@
...
+      };
+    });
+    return { ...data, top: result };
+  };
+
+  return (
+    <ReportingCard
+      text={{
+        title: 'Grower Gender Count',
+        text1: 'Total growers',
```
nmcharlton, line 38: Looks like there's no total in the data from the API. I'm not sure how useful it would be anyway.

src/components/reportingCards/ReportingCard.js
```diff
@@ -228,9 +236,15 @@ function GrowerReportingCard(props) {
                   <Icon className={classes.iconTotal} style={{ color }} />
                 </Box>
                 <Box className={classes.box3}>
-                  <Typography className={classes.total}>
-                    {new Intl.NumberFormat().format(data.num1)}
-                  </Typography>
+                  {typeof data.num1 === 'string' ? (
```
nmcharlton, line 239: As far as I can tell, all these stats come through from the API as strings and are formatted as locale-specific numbers using `Intl.NumberFormat().format()`.

nmcharlton: Same goes for the other places this has changed.

nmcharlton (changes requested):

src/components/reportingCards/ReportingCard8.js
gwynndp: Good call.

src/components/reportingCards/ReportingCard8.js
gwynndp: I agreed before, but now I think the total does end up being useful for interpreting the numbers of each gender.  

I added the total to the api data and made the gender_details more consistent with the other report cards' data now.  It saves us working around it on the client.

src/components/reportingCards/ReportingCard.js
gwynndp: I tried to simplify this now so that we can have the international number formats, and the percentage if it is included.

src/components/reportingCards/ReportingCard8.js
gwynndp: Percentages make more sense to me, but the format isn't great -- What do you think?

nmcharlton: Another option is to show both values with a separator, but we'd need to be careful with alignment.

Let's stick with raw numbers for now for consistency. It's easy for a user to convert numbers to percentages.

nmcharlton (approved):

Issue #754, opened 2022-07-19
Stakeholder tool: clear UX for assigning parents and child relationships


sebastiangaertner:
Is there a clear UX already for this? If not can we put this out there. There has been a couple of new UX designers that could take this on.

nmcharlton:
@sebastiangaertner Awesome! When a new stakeholder is created, the user can assign a parent, but there's no UX for subsequently linking two orgs.

https://user-images.githubusercontent.com/5558838/180306301-0233e610-8dab-4354-b844-f2687c5373ae.mov

The original proposition was a _Link Parent_ and a _Link Children_ button in the Stakeholder details dialog, but we didn't get as far as a design for what happens when you click it.
Clicking the _unlink_ icon next to a stakeholder should remove the link (once confirmed by the user).

From Figma:
![image](https://user-images.githubusercontent.com/5558838/180306217-7e3bd0e7-eaa2-4ea4-84a8-7ba5177b842a.png)

On a related note, our current UI suggests one stakeholder could have multiple parents, but we won't actually allow this. That means the _Parents_ container can be redesigned to show one parent.

Vlad-Pischaeff:
Dear sirs.
Please discuss getting data from endpoint for stackholders.

1. When we create a new holder, the data is sent in this format

![image](https://user-images.githubusercontent.com/48124219/186093894-9c329ffe-9021-4abd-82af-bf3f0fdc1623.png)

where the relation_id field is present

2. When we get the data from the server, they are already processed and the children do not have this field, and the parent has a field "parents", which he does not need at all

![image](https://user-images.githubusercontent.com/48124219/186094697-ed4dea6a-99f3-43df-b0f7-274cba1e18e3.png)

and in children the field"relation_id" has disappeared

![image](https://user-images.githubusercontent.com/48124219/186094945-de785e80-b2a3-43b3-9ca1-915a979e9bf0.png)

I think the server should send back raw data with all fields, and process it on the client side, or the clients save the field "relation_id" as an array. This will allow in the future to scale the application and simplify the client code

nmcharlton:
@gwynndp @Kpoke Are you able to respond to @Vlad-Pischaeff's proposal?

PR #792, opened 2022-08-23
fix(stakeholder tool): parent not showing #753

### LINKED ISSUE #753: Stakeholder tool: Parent not showing
Screenshot 2022-07-19 at 15 12 38

Vlad-Pischaeff:
## Description

_parent not showing #753_

**Issue(s) addressed**

- Resolves #753

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

**parent not showing**

**now parent show**

## Breaking change

**No breaking change**

### DIFF
--- a/src/components/Stakeholders/StakeholderDetail/Groups.js
+++ b/src/components/Stakeholders/StakeholderDetail/Groups.js
@@ -29,7 +29,7 @@ const useStyles = makeStyles({
   },
 });
 
-function StakeholderGroups({ data, render, tall }) {
+function StakeholderGroups({ data, render, tall, parentId }) {
   const classes = useStyles();
   const { stakeholders, setStakeholders } = useContext(StakeholdersContext);
 
@@ -50,6 +50,12 @@ function StakeholderGroups({ data, render, tall }) {
     setStakeholders(linked);
   };
 
+  const selectParent = (id) => {
+    return stakeholders.filter((item) => item.id === id);
+  };
+
+  const parent = selectParent(parentId);
+
   return (
     <>
       {render.map((type) => (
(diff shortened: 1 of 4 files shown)


src/components/Stakeholders/StakeholderDetail/Groups.js
```diff
@@ -64,7 +70,11 @@ function StakeholderGroups({ data, render, tall }) {
             <div className={classes.flex}>
               {stakeholder[type]?.icon}
               <Typography variant="h6">
-                {stakeholder[type]?.label} ({data[type]?.length || 0})
+                {stakeholder[type]?.label} (
+                {data[type]?.length ||
+                  (type === 'parents' && parent.length) ||
```
nmcharlton, line 75: Given we're doing multiple equality comparisons of `type`, it would be best to create some valid `const` or `enum` values rather than using string literals.

nmcharlton (changes requested):

Vlad-Pischaeff:
I suggest we put aside this PR, until we decide on my post https://github.com/Greenstand/treetracker-admin-client/issues/754#issuecomment-1223653644
If the child element has a field of "parent" type "array" with "relation_id" inside, then the solution may be simpler

src/components/Stakeholders/StakeholderDetail/Groups.js
Vlad-Pischaeff: I suggest we put aside this PR, until we decide on my post https://github.com/Greenstand/treetracker-admin-client/issues/754#issuecomment-1223653644
If the child element has a field of "parent" type "array" with "relation_id" inside, then the solution may be simpler

PR #794, opened 2022-08-23
Add species name to Verify tool capture tooltip 

### LINKED ISSUE #746: Add species name to Verify tool capture tooltip
Users need to quickly see what species (if any) has been assigned to captures in the Verify tool.
If `capture.speciesId` is set, species name should appear in the quick-look tooltip that appears when a user hovers over the preview card (`CaptureDetailTooltip`).

Species name is available as a lookup from `speciesContext.speciesList`.
The Material UI `Category` icon should be used.

Screenshot 2022-07-18 at 21 03 54

andres039:
## Description

On the CaptureDetailTooltip component, I have declared a `speciesName `variable that sets the name of the species on the tooltip as long as there is a capture.id property set. The name is looked up on the` speciesList ` which I am getting from the speciesContext (line 28) with the help of the `currentSpecies` function inside a useEffect (line 30). This useEffect has the `capture` variable as a dependency.

**Issue(s) addressed** 
   Add species name to Verify tool capture tooltip
- Resolves #746

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

   If there is a capture.id the tooltip doesn't show the Category icon nor the name of the species
**What is the new behavior?**
 When there is a capture.id set, the tooltip shows a Category icon and the name of the species
## Breaking change

No

### DIFF
--- a/src/components/CaptureDetailTooltip.js
+++ b/src/components/CaptureDetailTooltip.js
@@ -28,14 +28,14 @@ const CaptureDetailTooltip = ({ capture, showCaptureClick }) => {
   const { speciesList } = useContext(SpeciesContext);
 
   React.useEffect(() => {
-  setSpeciesName('');
+    setSpeciesName('');
     if (capture?.speciesId) {
       const currentSpecies = () => {
         return speciesList.find((species) => species.id === capture.speciesId);
       };
       const speciesData = currentSpecies();
       if (speciesData !== null) {
-      setSpeciesName(speciesData.name);
+        setSpeciesName(speciesData.name);
       }
     }
   }, [capture]);


src/components/CaptureDetailTooltip.js
```diff
@@ -21,9 +24,20 @@ const CaptureDetailTooltip = ({ capture, showCaptureClick }) => {
...
+  const { speciesList } = useContext(SpeciesContext);
 
-  const CaptureDetailTooltipStyles = CaptureDetailTooltipUseStyles();
+  React.useEffect(() => {
+    if (capture?.speciesId) {
+      const currentSpecies = () => {
+        return speciesList.find((species) => species.id === capture.speciesId);
+      };
+      const speciesData = currentSpecies();
+      setSpeciesName(speciesData.name);
```
nmcharlton, line 36: Need to check that `speciesData` is not `null` first.

nmcharlton (changes requested):
Looks great! Just one minor comment.

src/components/CaptureDetailTooltip.js
andres039: Thank you, that makes sense. I'll try to fix it.

nmcharlton: We also need to clear the species name if there is no match, otherwise it may display the previous capture's species name.

andres039: Thank you very much, I'm still learning the clean-up functions and it was a good learning experience. I think I got  it right this time.

src/components/CaptureDetailTooltip.js
```diff
@@ -21,9 +24,23 @@ const CaptureDetailTooltip = ({ capture, showCaptureClick }) => {
...
+    if (capture?.speciesId) {
+      const currentSpecies = () => {
+        return speciesList.find((species) => species.id === capture.speciesId);
+      };
+      const speciesData = currentSpecies();
+      if (speciesData !== null) {
+      setSpeciesName(speciesData.name);
+      }
+    }
+    return () => setSpeciesName();
```
nmcharlton, line 40: This isn't quite what I had in mind. I was thinking more like the following:
```js
React.useEffect(() => {
  setSpeciesName(''); // <- Just clear speciesName before trying to retrieve the species
  if (capture?.speciesId) {
    const currentSpecies = () => {
      return speciesList.find((species) => species.id === capture.speciesId);
    };
    const speciesData = currentSpecies();
    if (speciesData !== null) {
      setSpeciesName(speciesData.name);
    }
  }
}
```

nmcharlton (approved):
Thanks @andres039!

PR #826, opened 2022-09-08
feat(regions): add download button, onClick returns a downloadable ge…

### LINKED ISSUE #777: Regions: Allow users to download geojson shapefiles
Add a _download_ button to records in the Regions table.
The _download_ icon can be inserted before the _edit_ and _delete_ icons.

The endpoint `REGION_API/region/:region_id/shape` returns a downloadable geojson file.

andres039:
…ojson file

## Description
I added a download icon before the edit button on the regions tab.
I created a function that does a Get request to the endpoint. Since I am using fetch, I check for the` response.ok` property before allowing the function to redirect to the endpoint.  

**Issue(s) addressed**
Regions: Allow users to download geojson shapefiles
- Resolves #777

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

There is no icon to download the geojson file
**What is the new behavior?**
There is an icon and a function that allow for the geojson to be downloaded onClick
## Breaking change
Screenshot from 2022-09-08 10-33-20
Screenshot from 2022-09-08 10-32-51

### DIFF
--- a/src/components/Regions.js
+++ b/src/components/Regions.js
@@ -28,14 +28,7 @@ import {
   Snackbar,
 } from '@material-ui/core';
 import { ToggleButtonGroup, ToggleButton } from '@material-ui/lab';
-import {
-  Edit,
-  Close,
-  CloudDownload,
-  GetApp,
-  Delete,
-  Map,
-} from '@material-ui/icons';
+import { Edit, Close, GetApp, Delete, Map } from '@material-ui/icons';
 import { withStyles } from '@material-ui/core/styles';
 import { RegionContext } from '../context/RegionContext';
 import { AppContext } from '../context/AppContext';


src/components/Regions.js
```diff
@@ -247,6 +265,9 @@ const RegionTable = (props) => {
           </>
         )}
         <TableCell align="right" className={classes.operations}>
+          <IconButton title="download" onClick={() => handleDownload(item)}>
+            <CloudDownload />
```
nmcharlton, line 269: For consistency with the _Earnings_ tool, let's use the `Download` icon. (_Earnings_ actually uses `GetApp`, but that's just an alias for `Download`).

nmcharlton (changes requested):
Excellent work @andres039!
Just one minor suggestion.

src/components/Regions.js
andres039: Thank you @nmcharlton. There is a new commit where the icon has been changed. I appreciated your clarification about the alias, I simply couldn't find the icon in the node modules but, I did learn something new.

src/components/Regions.js
```diff
@@ -266,7 +273,7 @@ const RegionTable = (props) => {
         )}
         <TableCell align="right" className={classes.operations}>
           <IconButton title="download" onClick={() => handleDownload(item)}>
-            <CloudDownload />
+            <GetApp />
```
nmcharlton, line 269: Sorry to be awkward; can we use `Download` (https://mui.com/material-ui/material-icons/?selected=Download), rather than `GetApp`? As you've seen, `GetApp` can be confusing when you're browsing the code.

It'd be great if you can update the instances in _CustomTable.js_ and _CaptureTable.js_ as well.

andres039: That makes sense. However, is it OK to install a new package? I think I need to install @mui/icons-material for this icon to work. And of course, I'll update _CutomTable.js_ and _CaptureTable.js_

nmcharlton: Sorry, that's my fault. I was looking at v5 Material UI, which we're not using yet.
Let's stick with `GetApp` until we've completed migration (under a separate issue).

PR #832, opened 2022-09-18
feat: update verify approve/reject links to point to new microservices

gwynndp:
## Description

- as a first step to migrating from the legacy api to the new microservices, update the verify approve/reject process to update the field_data api and treetracker api 
- verify still pulls data from the legacy tables
- a field-data-service will need to update the legacy tables to keep the approved captures in sync
- the field_data raw_captures table will also need to have the raw_captures migrated from the legacy tree table
- the treetracker tags table will also need to have legacy tags migrated to work exactly the same, but this isn't required if we don't care about exact matches with current test data

**Issue(s) addressed**

- Resolves #issue_number

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/api/treeTrackerApi.js
+++ b/src/api/treeTrackerApi.js
@@ -1,65 +1,62 @@
 import { handleResponse, handleError, getOrganization } from './apiUtils';
+import { getVerificationStatus } from '../common/utils';
 import { session } from '../models/auth';
 import log from 'loglevel';
 
 // Set API as a variable
 const API_ROOT = process.env.REACT_APP_API_ROOT;
 const FIELD_DATA_API = process.env.REACT_APP_FIELD_DATA_ROOT;
+const QUERY_API = process.env.REACT_APP_QUERY_API_ROOT;
 const TREETRACKER_API = process.env.REACT_APP_TREETRACKER_API_ROOT;
-
-const CAPTURE_FIELDS = {
-  uuid: true,
-  imageUrl: true,
-  lat: true,
-  lon: true,
-  id: true,
-  timeCreated: true,
-  timeUpdated: true,
-  active: true,
-  approved: true,
-  planterId: true,
-  deviceIdentifier: true,
-  planterIdentifier: true,
-  speciesId: true,
-  tokenId: true,
-  morphology: true,
-  age: true,
-  captureApprovalTag: true,
-  rejectionReason: true,
-  note: true,
+const STATUS_STATES = {
+  Approved: 'approved',
+  Rejected: 'rejected',
+  'Awaiting Verification': 'unprocessed',
 };
 
 export default {
+  makeQueryString(filterObj) {
+    let arr = [];
+    for (const key in filterObj) {
+      if ((filterObj[key] || filterObj[key] === 0) && filterObj[key] !== '') {
+        arr.push(`${key}=${filterObj[key]}`);
+      }
+    }
+
+    return arr.join('&');
+  },
   /**
    * Verify Tool
    */
-  getCaptureImages(
+  getRawCaptures(
     {
-      skip,
+      page = 0,
       rowsPerPage,
-      orderBy = 'id',
-      order = 'desc',
-      //the filter model
+      // TODO: need to be implemented with field data API
+      // orderBy = 'captured_at',
+      // order = 'desc',
       filter,
     },
     abortController
   ) {
     try {
       const where = filter.getWhereObj();
+      where.status =
+        STATUS_STATES[getVerificationStatus(where.active, where.approved)];
+      delete where.active;
+      delete where.approved;
 
-      const filterData = {
-        where,
-        order: [`${orderBy} ${order}`],
+      log.debug('getRawCaptures where -->', where);
+
+      const filterObj = {
+        ...where,
         limit: rowsPerPage,
-        skip,
-        fields: CAPTURE_FIELDS,
+        offset: page * rowsPerPage,
       };
 
-      log.debug('getCaptureImages', filterData);
-
-      const query = `${API_ROOT}/api/${getOrganization()}trees?filter=${JSON.stringify(
-        filterData
-      )}`;
+      const query = `${QUERY_API}/raw-captures${
+        filterObj ? `?${this.makeQueryString(filterObj)}` : ''
+      }`;
 
       return fetch(query, {
         headers: {
@@ -71,7 +68,7 @@ export default {
       handleError(error);
     }
   },
-  approveCaptureImage(capture, morphology, speciesId, captureApprovalTag) {
+  approveCaptureImage(capture, morphology, age, captureApprovalTag, speciesId) {
     try {
       log.debug('approveCaptureImage', capture, captureApprovalTag);
 
@@ -90,24 +87,10 @@ export default {
         gps_accuracy: capture.gps_accuracy,
         captured_at: capture.timeCreated,
         note: capture.note,
+        age: age,
         morphology,
-        species_id: speciesId, // need uuid
-        // captureApprovalTag, // how does this fit into the new API?
+        species_id: speciesId,
       };
-
-      log.debug('newCapture data', newCapture);
-
-      // update the raw capture
-      fetch(`${FIELD_DATA_API}/raw-captures/${capture.uuid}`, {
-        method: 'PATCH',
-        headers: {
-          'content-type': 'application/json',
-          Authorization: session.token,
-        },
-        body: JSON.stringify({
-          status: 'approved',
-        }),
-      }).then(handleResponse);
 
       // add the new capture
       fetch(`${TREETRACKER_API}/captures`, {
@@ -122,9 +105,9 @@ export default {
       handleError(error);
     }
   },
-  rejectCaptureImage(capture, rejection_reason) {
+  rejectCaptureImage(capture, rejectionReason) {
     try {
-      log.debug('reject capture', capture.uuid, rejection_reason);
+      log.debug('reject capture', capture.uuid, rejectionReason);
       const query = `${FIELD_DATA_API}/raw-captures/${capture.uuid}`;
       return fetch(query, {
         method: 'PATCH',
@@ -134,7 +117,7 @@ export default {
         },
         body: JSON.stringify({
           status: 'rejected',
-          rejection_reason,
+          rejection_reason: rejectionReason,
         }),
       }).then(handleResponse);
     } catch (error) {
@@ -206,14 +189,15 @@ export default {
       handleError(error);
     }
   },
-  getCaptureById(id) {
+  getCaptureById(url, id, abortController) {
     try {
-      const query = `${API_ROOT}/api/${getOrganization()}trees/${id}`;
-
+      // use field data api for Verify and query api for Captures
+      const query = `${url}/${id}`;
       return fetch(query, {
         headers: {
           Authorization: session.token,
         },
+        signal: abortController?.signal,
       }).then(handleResponse);
     } catch (error) {
       handleError(error);
@@ -376,9 +360,10 @@ export default {
    */
   getTags(abortController) {
     try {
+      // TODO: order is not allowed as a filter
       // const filterString = `order=name`;
       // const query = `${TREETRACKER_API}/tags?${filterString}`;
-      const query = `${TREETRACKER_API}/tags`; // TODO: order is not allowed as a filter
+      const query = `${TREETRACKER_API}/tags`;
 
       return fetch(query, {
         method: 'GET',
@@ -395,7 +380,6 @@ export default {
   getTagById(id) {
     try {
       const query = `${TREETRACKER_API}/tags/${id}`;
-      //     const query = `${API_ROOT}/api/tags/${id}`;
       return fetch(query, {
         method: 'GET',
         headers: {
@@ -442,7 +426,7 @@ export default {
       handleError(error);
     }
   },
-  async getCaptureTags(captureIds = []) {
+  async getCaptureTags({ captureIds = [] }) {
     try {
       const result = captureIds.map((id) => {
         const query = `${TREETRACKER_API}/captures/${id}/tags`;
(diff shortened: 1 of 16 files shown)


gwynndp:
@Kpoke Tagging you to keep you in the loop on the client side.  This is NOT pulling from the microservice api but it is posting changes to the treetracker and field_data apis.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+        id: capture.uuid,
+        reference_id: capture.id,
+        session_id: capture.session_id, // no legacy equivalent
+        grower_account_id: capture.planterId,
+        planting_organization_id: capture.organization_id,
+        device_configuration_id: capture.device_configuration_id, // no legacy equivalent
+        image_url: capture.imageUrl,
+        lat: capture.lat,
+        lon: capture.lon,
+        gps_accuracy: capture.gps_accuracy,
```
nmcharlton, line 95: I would expect this to come through from LoopBack as `gpsAccuracy`, although it's currently omitted in `CAPTURE_FIELDS`. For simplicity, I'd suggest removing the explicit field definition from the `getCaptureImages()` query and just getting everything.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+        captured_at: capture.timeCreated,
+        note: capture.note,
+        morphology,
+        species_id: speciesId, // need uuid
```
nmcharlton, line 99: Users currently use Verify to add a species or some other tag to an already approved capture. We need to make sure we don't discard the existing `capture.speciesId` if one exists and the user hasn't set another (this highlights the need to make the UX clearer in this use case).

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
       handleError(error);
     }
   },
-  approveCaptureImage(id, morphology, age, captureApprovalTag, speciesId) {
```
nmcharlton, line 71: Let's keep `age` for now until we get round to refactoring the UI. We don't want to drop any information the user has entered.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+        device_configuration_id: capture.device_configuration_id, // no legacy equivalent
+        image_url: capture.imageUrl,
+        lat: capture.lat,
+        lon: capture.lon,
+        gps_accuracy: capture.gps_accuracy,
+        captured_at: capture.timeCreated,
+        note: capture.note,
+        morphology,
+        species_id: speciesId, // need uuid
+        captureApprovalTag, // how does this fit into the new API?
```
nmcharlton, line 100: I see an `attributes` field. Could that be it?
We may want to shift this tag to just be another `capture_tag`, but one that uses preset options.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+          Authorization: session.token,
+        },
+        body: JSON.stringify(newCapture),
+      }).then(handleResponse);
     } catch (error) {
       handleError(error);
     }
   },
-  rejectCaptureImage(id, rejectionReason) {
+  rejectCaptureImage(capture, rejection_reason) {
```
nmcharlton, line 130: Is there a reason the `rejection_reason` parameter is snake case?
It's best to keep use of camel case consistent within the Admin Panel and only convert when populating the API body (`rejection_reason: rejectionReason`).

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
         body: JSON.stringify({
-          id: id,
-          approved: true,
-          //revise, if click approved on a rejected pic, then, should set the pic approved, AND restore to ACTIVE = true
-          active: true,
-          morphology,
-          age,
-          captureApprovalTag,
-          speciesId: speciesId,
+          status: 'approved',
```
nmcharlton, line 113: I'd prefer to see these statuses defined as a const enum somewhere, rather than as string literals in the code.
That way it's easier to make sure they're in sync with the API.

src/api/treeTrackerApi.js
```diff
@@ -371,8 +399,8 @@ export default {
   },
   getTagById(id) {
     try {
-      const query = `${API_ROOT}/api/tags/${id}`;
-
+      const query = `${TREETRACKER_API}/tags/${id}`;
+      //     const query = `${API_ROOT}/api/tags/${id}`;
```
nmcharlton, line 403: We can remove this.

src/api/treeTrackerApi.js
```diff
@@ -407,28 +431,43 @@ export default {
   /*
    * Capture Tags
    */
-  createCaptureTags(captureId, tags) {
+  createCaptureTags(capture_id, tags) {
```
nmcharlton, line 434: Again, I'd prefer to use camel case for function parameters.

src/context/TagsContext.js
```diff
@@ -25,17 +26,31 @@ export function TagsProvider(props) {
   // EVENT HANDLERS
   const loadTags = async () => {
     const tags = await api.getTags();
```
nmcharlton, line 28: I'd suggest renaming this to `result` or `response` so that we don't end up with `tags.tags`.

src/context/TagsContext.js
```diff
@@ -25,17 +26,31 @@ export function TagsProvider(props) {
...
   /*
    * check for new tags in tagInput and add them to the database
    */
+
   const createTags = async () => {
+    const orgId = getOrganizationUUID();
+    log.debug('orgId:', orgId);
+    const newTagTemplate = {
+      isPublic: orgId ? false : true,
+      owner_id: orgId,
```
nmcharlton, line 41: Just a thought, we will need a way to verify that the user has permission to set the `owner_id` to a particular org on the backend (using our Keycloak integration).

src/context/VerifyContext.js
```diff
@@ -102,26 +102,26 @@ export function VerifyProvider(props) {
 
   // EVENT HANDLERS
 
-  const approve = async ({ approveAction, id }) => {
+  const approve = async ({ approveAction, captureImage }) => {
```
nmcharlton, line 105: I'd suggest we remove the `Image` part from `captureImage` and `api.approveCaptureImage`. We're approving the whole capture, not just the image, so the names are a bit confusing.

src/context/VerifyContext.js
```diff
@@ -102,26 +102,26 @@ export function VerifyProvider(props) {
...
     if (!approveAction) {
       throw Error('no approve action object!');
     }
     if (approveAction.isApproved) {
       log.debug('approve');
       await api.approveCaptureImage(
```
nmcharlton, line 111: These inputs no longer tally with the updated function signature.

nmcharlton (changes requested):

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
 
-      log.debug(query);
+      // map legacy data to fields for new microservice
+      const newCapture = {
+        id: capture.uuid,
+        reference_id: capture.id,
+        session_id: capture.session_id, // no legacy equivalent
+        grower_account_id: capture.planterId,
+        planting_organization_id: capture.organization_id,
+        device_configuration_id: capture.device_configuration_id, // no legacy equivalent
```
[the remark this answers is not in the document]
gwynndp: For the moment I'm trusting that Daniels field-data-service will preserve that data by making the connections with the correct raw_capture for the microservices and update the right tree in the legacy tree table.

src/api/treeTrackerApi.js
gwynndp: I tried that first, but the treetracker api will reject an object with fields that are "not allowed".  I can comment out those fields but some of them are required, so I decided that there would still be work to do but allowing some undefined values for the moment would let us see how it fits with the field-service-api and exactly what would need to be changed.

src/api/treeTrackerApi.js
gwynndp: Noted.

src/context/TagsContext.js
gwynndp: Good thought --- I'm making the temporary assumption that keycloak will manage access via permissions and that the user will only have access to approve/reject the organizations that they are allowed to edit.  The tags are currently added to a general table, but they are associated to a particular capture via id on a join table.  So a user can add tags to a capture when they approve it, but they can't edit them directly or without going through a capture that belongs to their approved organization.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+      );
 
-      log.debug(query);
+      // map legacy data to fields for new microservice
+      const newCapture = {
+        id: capture.uuid,
+        reference_id: capture.id,
+        session_id: capture.session_id, // no legacy equivalent
+        grower_account_id: capture.planterId,
+        planting_organization_id: capture.organization_id,
```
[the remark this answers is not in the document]
gwynndp: There isn't actually a plantingOrganizationId on the capture objects in Verify.  I'll see if I can pull it from the current filter and if it isn't there I'll get it from the logged in organization.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+        captureApprovalTag
+      );
 
-      log.debug(query);
+      // map legacy data to fields for new microservice
+      const newCapture = {
+        id: capture.uuid,
+        reference_id: capture.id,
+        session_id: capture.session_id, // no legacy equivalent
+        grower_account_id: capture.planterId,
```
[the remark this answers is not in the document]
gwynndp: You're correct, it won't be handled by the treetracker api, but I think it would be needed by the field-data-service to identify the correct grower and the grower_account_id could be added that way.  @Kpoke Your thoughts?

src/api/treeTrackerApi.js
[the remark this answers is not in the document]
gwynndp: @Kpoke Thoughts?

src/context/VerifyContext.js
gwynndp: Do you mean I should update the naming of approveAction and captureImage?  I updated from id where approve is invoked.  Verify seems to refer to the capture object as captureImage throughout.  Sorry, I don't understand what change is needed.

src/context/VerifyContext.js
gwynndp: Agreed.  I'll change it.

gwynndp: I read the below comment first, but I think this clarifies it.

src/api/treeTrackerApi.js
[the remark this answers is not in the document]
Kpoke: The field-data service already has that information. The wallet_registrations table has that column populated.

src/api/treeTrackerApi.js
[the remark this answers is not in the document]
Kpoke: First the treetracker.capture is currently empty so there isn't a case of a capture without a device_configuration_id, also both the treetracker and field-data would be updating the trees in the legacy table. Then, the device_identifier and the device uuid are connected for data coming from v1, the uuid is generated from the device_identifier

src/api/treeTrackerApi.js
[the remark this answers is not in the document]
nmcharlton: It's only missing because we're specifying a subset of fields in the API query. It should be available to query.

src/api/treeTrackerApi.js
nmcharlton: Sorry, you're right – it's in the legacy `trees` table, but not in the LoopBack API.
Similarly to some of the other fields, we can either update the legacy API to include it (undesirable) or have a migration get the missing fields.

src/context/VerifyContext.js
nmcharlton: Here you're calling:
```
approveCaptureImage(capture, morphology, age, captureApprovalTag, speciesId)
```
but the function signature in `treeTrackerApi.js` is:
```
approveCaptureImage(capture, morphology, speciesId, captureApprovalTag)
```

I'd suggest reverting the one in `treeTrackerApi.js` to get `age` back in.

src/context/VerifyContext.js
nmcharlton: I think it's worth replacing `captureImage` with `capture` throughout Verify to make its meaning clearer.

src/context/VerifyContext.js
gwynndp: Thank you for the explanation.  I was looking at the wrong function.  Will do.

src/api/treeTrackerApi.js
```diff
@@ -68,51 +71,75 @@ export default {
...
+        morphology,
+        species_id: speciesId, // need uuid
+        captureApprovalTag, // how does this fit into the new API?
+      };
 
-      return fetch(query, {
+      log.debug('newCapture data', newCapture);
+
+      // update the raw capture
+      fetch(`${FIELD_DATA_API}/raw-captures/${capture.uuid}`, {
```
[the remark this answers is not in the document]
gwynndp: @Kpoke I thought we needed to make both calls in order to pass along the rejection reason and approval tags to the apis that need them.  The treetracker api is only for capture approvals, not rejections, but field-data needs to be updated for both approvals and rejections.

src/api/treeTrackerApi.js
gwynndp: @Kpoke and I discussed how to handle some of these discrepancies on Tues and figured that we'd best fetch the field-data raw-captures to Verify so that the data fields align.  It's the direction we're moving anyway and it should let each tool mostly work. 

There will be conflicts with data that crosses the legacy and new apis like GrowerDetail and Capture Detail -- the data won't always be found since we're still in progress merging it to the new microservice.

src/api/treeTrackerApi.js
gwynndp: I had started naming the parameters to match the keys in the data before I got your message about preferring camelcase parameters.  Will switch.

nmcharlton (approved):

PR #854, opened 2022-09-30
feat: fix filter search contains all

### LINKED ISSUE #686: Fix way of filtering for "All" values in Earnings Table filter
This is a risky way of spotting that all values should be included.
If a user wants to filter by a grower name that starts with or contains "all", we will ignore the value and return all growers.

_Originally posted by @nmcharlton in https://github.com/Greenstand/treetracker-admin-client/pull/625#discussion_r889703220_

Hengxin-Liu:
## Description

**Issue(s) addressed**
If a user wants to filter by a grower name that starts with or contains "all", we will
 ignore the value and return all growers.

- Resolves #686 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change


src/components/EarningsTable/EarningsTable.js
```diff
@@ -8,16 +8,7 @@ import {
 } from 'utilities';
 import CustomTableFilter from 'components/common/CustomTableFilter/CustomTableFilter';
 import CustomTableItemDetails from 'components/common/CustomTableItemDetails/CustomTableItemDetails';
-/**
```
nmcharlton, line 11: What's the reason for removing this block?

src/components/EarningsTable/EarningsTable.js
```diff
@@ -140,7 +131,7 @@ function EarningsTable() {
     // filter out keys we don't want to submit
     Object.keys(filtersToSubmit).forEach((k) => {
       if (
-        filtersToSubmit[k] === 'all' ||
+        filtersToSubmit[k].includes('all') ||
```
nmcharlton, line 134: Unfortunately this causes the opposite behaviour to what we want.
User should be able to type "all" into the filter and see only results containing the string "all".
We need another way to represent _show all_ in `filtersToSubmit`.

nmcharlton (changes requested):

src/components/EarningsTable/EarningsTable.js
Hengxin-Liu: I seems like I accidentally deleted it, I fix it now

src/components/EarningsTable/EarningsTable.js
nmcharlton: @Hengxin-Liu Are you still working on this issue?

PR #857, opened 2022-10-01
fix: removed download button from regions page

### LINKED ISSUE #850: Regions: Remove download button from Collections
The download button against Collection records does not work and should be removed.
Screenshot 2022-09-28 at 21 35 03

dharmiknotop:
## Description

Removed the download button from the regions page

**Issue(s) addressed**

- Resolves #850

- [ ] Bug fix

## Issue

The Download button doesn't work and no need of it there.

**What is the new behavior?**
removed the download button

### DIFF
--- a/src/components/Regions.js
+++ b/src/components/Regions.js
@@ -265,6 +265,11 @@ const RegionTable = (props) => {
           </>
         )}
         <TableCell align="right" className={classes.operations}>
+          {!showCollections && (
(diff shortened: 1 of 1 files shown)


nmcharlton (changes requested):
We need to keep the download button when the Regions switch is selected, just remove it for Collections.

nmcharlton (approved):
Nice work @dharmiknotop

PR #895, opened 2022-10-14
Remove id columns from species table

### LINKED ISSUE #875: Add uuid to species table and api response

gwynndp:
## Description

- the current table displays the species id but we want don't want that to be public information, just an internal identifier

**What is the new behavior?**
- remove the id column

**Issue(s) addressed**

- Resolves #875

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

### DIFF
--- a/src/components/SpeciesTable.js
+++ b/src/components/SpeciesTable.js
@@ -210,9 +210,6 @@ const SpeciesTable = (props) => {
             checked={selected.includes(species.id)}
           />
         </TableCell>
-        <TableCell component="th" scope="row">
-          {species.id}
-        </TableCell>
         <TableCell component="th" scope="row" data-testid="species">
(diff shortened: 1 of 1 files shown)


nmcharlton:
@gwynndp Is there a need for the ID (UUID) in the table? Since the ID is an internal identifier, it doesn't really serve a purpose in the UI.

gwynndp:
> 

My reasoning is that since we were showing the the ID and that will be replaced by the UUID we should add it to the management view.  Perhaps we don't need any id displayed in the Species table then.  I can remove both ID and UUID.

nmcharlton:
Agreed. Let's just remove the ID from the table and leave the UUID off.

nmcharlton (approved):

PR #922, opened 2022-11-06
Display notes under captures' images

### LINKED ISSUE #667: Notes field should be visible for every tree in the capture matching tool (maybe even in other views)
We have seen that the "notes" field in the app is often used to define if the capture is a first tracking or maintenance visit, therefore  it would make sense to show the content of the notes field in the capture matching tool. this could even be done with some kind of sticky note graphic and the notes content on that right on the image. https://media.istockphoto.com/photos/yellow-note-paper-picture-id459886461 as a sample. We can also extend the opacity based bottom part with the date and distance already displayed. 

In case of the capture matching tool this would give context that sticks out and can be used. Additionally the grower (and org or later contract) could be shown on the sticky note.

janet-yu:
## Description
Display a capture's notes underneath its image. Let me know if there are any other design considerations to take into account!

**Issue(s) addressed**

- Resolves #667 

- [X] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

Notes are not displayed under captures.
**What is the new behavior?**
Notes are displayed.
## Breaking change

## Other useful information
I didn't want it to be too complicated, so I ended up just adding the notes under each capture:
notes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# [1.114.0](https://github.com/Greenstand/treetracker-admin-client/compare/v1.113.0...v1.114.0) (2022-11-10)
+
+
+### Features
+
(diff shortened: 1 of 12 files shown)


dadiorchen:
Awsome, this is getting more and more important for us

nmcharlton (approved):
Nice work @janet-yu!

PR #931, opened 2022-11-12
Fix displaying capture location and Google maps query

### LINKED ISSUE #930: The location and country name with capture matching tool is broken
As shown in the picture, the location of the capture is empty, and if click the location icon, the map is wrong to show the location of this capture.
Screen Shot 2022-11-11 at 12 00 19 PM

janet-yu:
## Description
Make sure to check for `lat` and `lon` properties on the capture when displaying location details.

**Issue(s) addressed**
- Resolves #930 

- [ ] Enhancement
- [X] Bug fix
- [ ] Refactor

## Issue

Country is not displayed above capture and clicking the location marker leads to an undefined location.

**What is the new behavior?**
Country is displayed properly and redirected to correct location in Google maps.

## Breaking change

No

### DIFF
--- a/src/components/CaptureMatching/CaptureMatchingView.js
+++ b/src/components/CaptureMatching/CaptureMatchingView.js
@@ -417,12 +417,17 @@ function CaptureMatchingView() {
     setIsDetailsPaneOpen(false);
   };
 
-  const countryInfo = useMemo(
-    () => (
-      <Country lat={captureImage?.latitude} lon={captureImage?.longitude} />
-    ),
-    [captureImage?.latitude, captureImage?.longitude]
-  );
+  const countryInfo = useMemo(() => {
+    const latitude = captureImage?.latitude || captureImage?.lat;
+    const longitude = captureImage?.longitude || captureImage?.lon;
+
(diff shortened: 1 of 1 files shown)


nmcharlton (commented):
Using a truthy check like this on a lat or lon value will return `false` if the value is valid but exactly zero. That's very unlikely but not impossible, so checking on whether the value is a number might be best.

janet-yu:
Ah, good point! So if the lat and lon are are both 0, it looks like `Country` will return "can not find country at lat: 0,  lon: 0" and that's displayed as the location. Do we want to keep this behavior? If so, maybe we can keep these checks as is.
cannotfind

nmcharlton:
If _both_ `lat` and `lon` are 0, it's pretty safe to assume we have invalid coordinates.
We just need to allow for situations where one of the values is 0.

janet-yu:
I see - I think we should be okay with this check though, since the latitude and longitudes are strings. If any of the values are "0", that should be truthy (I tested scenarios where one value is "0", and it returns the country). Let me know if I'm misunderstanding something!

nmcharlton (approved):

PR #963, opened 2022-11-27
feat: update captures to use treetracker microservice

### LINKED ISSUE #935: Captures: AxiosError: Internal Server Error
On the Captures page, an error pops up in the console when submitting a Capture UUID  - 500 (Internal Server Error).

Screenshot 2022-11-14 at 1 39 08 PM

gwynndp:
## Description

- update api queries to reference new Query api and Treetracker api
- update filter to match the new data: remove verification status filter, change to Wallet filter from Grower Identifier
- display reference id instead of capture uuid to user
- update Filter Model to match new filter and api data structure

**Issue(s) addressed**

- Resolves #935

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

Screen Shot 2022-12-22 at 2 28 38 PM

### DIFF
--- a/src/components/CaptureDetailDialog.js
+++ b/src/components/CaptureDetailDialog.js
@@ -113,27 +113,7 @@ function CaptureDetailDialog({ open, captureId, onClose, page }) {
     const current = cdContext.capture;
     if (current) {
       // map the keys from legacy to new api keys
-      setRenderCapture({
-        status: current.status,
-        id: current.id || current.uuid,
-        reference_id: current.reference_id || current.id,
-        grower_account_id: current.grower_account_id || current.planterId,
-        wallet: current.wallet || current.planterIdentifier,
-        device_identifier:
-          current.device_identifier || current.deviceIdentifier,
-        image_url: current.image_url || current.imageUrl,
-        lat: current.lat || current.latitude,
-        lon: current.lon || current.longitude,
-        age: current.age,
-        captureApprovalTag: current.captureApprovalTag || null,
-        morphology: current.morphology || null,
-        note: current.note,
-        rejectionReason: current.rejectionReason || null,
-        species_id: current.species_id || current.speciesId,
-        token_id: current.token_id || current.tokenId,
-        created_at: current.created_at || current.timeCreated,
-        updated_at: current.updated_at || current.timeUpdated,
-      });
+      setRenderCapture({ ...current });
     }
     if (isLoading) {
       setIsLoading(false);
@@ -277,12 +257,14 @@ function CaptureDetailDialog({ open, captureId, onClose, page }) {
           <Typography className={classes.subtitle}>
             Verification Status
           </Typography>
-          {capture.status === captureStatus.UNPROCSSED ? (
+          {capture.status === captureStatus.UNPROCESSED ? (
             <Chip
               label={verificationStates.AWAITING}
               className={classes.awaitingChip}
             />
-          ) : capture.status === captureStatus.APPROVED ? (
+          ) : // Verify will have status of 'approved', Captures and CaptureMatch will have status of 'active' because all captures are approved
+          capture.status === captureStatus.APPROVED ||
+            capture.status === 'active' ? (
             <Chip
               label={verificationStates.APPROVED}
               className={classes.approvedChip}
--- a/src/components/CaptureFilter.js
+++ b/src/components/CaptureFilter.js
@@ -74,20 +74,17 @@ function Filter(props) {
   const tagsContext = useContext(TagsContext);
   const { classes, filter = new FilterModel() } = props;
   const filterOptionAll = 'All';
-  const dateStartDefault = null;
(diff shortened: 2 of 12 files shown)


src/api/treeTrackerApi.js
```diff
@@ -18,7 +18,26 @@ export default {
   makeQueryString(filterObj) {
     let arr = [];
     for (const key in filterObj) {
-      if ((filterObj[key] || filterObj[key] === 0) && filterObj[key] !== '') {
+      if (key === 'device_identifier') {
```
nmcharlton, line 21: Use [`encodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) instead.
Assuming the API supports it, it would be a good idea to do this with all `string` filters.

src/components/CaptureDetailDialog.js
```diff
@@ -288,7 +287,7 @@ function CaptureDetailDialog(props) {
               label={verificationStates.AWAITING}
               className={classes.awaitingChip}
             />
-          ) : capture.status === 'approved' ? (
+          ) : capture.status === 'active' || capture.status === 'approved' ? (
```
nmcharlton, line 290: What's this `active` status?
Also, I think you know what I'm going to say about using string literals for comparison 😜 - best to define them once as a `const` object and reference that.

src/components/CaptureFilter.js
```diff
@@ -338,7 +252,7 @@ function Filter(props) {
               <TextField
                 htmlFor="grower-identifier"
                 id="grower-identifier"
-                label="Grower Identifier"
+                label="Wallet"
                 placeholder="e.g. grower@example.com"
                 value={growerIdentifier}
```
nmcharlton, line 257: Does this need to become `wallet`?

src/components/CaptureFilter.js
```diff
@@ -306,23 +220,23 @@ function Filter(props) {
               <TextField
                 htmlFor="grower-id"
                 id="grower-id"
-                label="Grower ID"
+                label="Grower Account ID"
                 placeholder="e.g. 2, 7"
                 value={growerId}
```
nmcharlton, line 225: Can we change the names of these variables to match the new microservice schema?

src/context/CapturesContext.js
```diff
@@ -69,44 +60,49 @@ export function CapturesProvider(props) {
...
-    return response;
+    api
+      .getCaptureById(`${process.env.REACT_APP_QUERY_API_ROOT}/v2/captures`, id)
+      .then((res) => {
+        setIsLoading(false);
+        setCapture(res);
+      })
+      .catch((err) => {
+        setIsLoading(false);
+        console.error(`ERROR: FAILED TO GET SELECTED TREE ${err}`);
```
nmcharlton, line 104: Use `loglevel` rather than `console.error`, and it's a capture rather than a tree 😄

src/models/Filter.js
```diff
@@ -9,24 +9,30 @@ export const ALL_ORGANIZATIONS = 'ALL_ORGANIZATIONS';
...
-  dateEnd;
-  approved;
-  active;
-  planterId;
-  deviceIdentifier;
-  planterIdentifier;
-  speciesId;
+  dateStart; //  startDate
+  dateEnd; //  endDate
+  approved; //  status...
```
nmcharlton, line 19: Can we do away with `approved` and `active` now?

src/models/Filter.js
```diff
@@ -9,24 +9,30 @@ export const ALL_ORGANIZATIONS = 'ALL_ORGANIZATIONS';
...
   captureId;
-  dateStart;
-  dateEnd;
-  approved;
-  active;
-  planterId;
-  deviceIdentifier;
-  planterIdentifier;
-  speciesId;
+  dateStart; //  startDate
```
nmcharlton, line 17: Let's rename all of these throughout the codebase.

nmcharlton (changes requested):

src/api/treeTrackerApi.js
gwynndp: Oh, nice! Thanks

src/components/CaptureFilter.js
gwynndp: I figure we're going to get confused if we don't call it the same thing in the UX as we do in the database, but the rest didn't have to change because it is only visible in the code.  Eventually we can change all the form references if needed.

src/components/CaptureFilter.js
gwynndp: Yes, I think we should, but didn't want to do it until we are fully switched over.  Maybe that can be a cleanup issue after we confirm everything is working?  Unless you think it's a good idea to do now.

nmcharlton: I'm happy to do this under a cleanup issue.

PR #974, opened 2022-12-07
feat: add wallet filter to growers

### LINKED ISSUE #952: Add wallet filter to Growers
Currently there is no way to filter the growers by their wallet 

Wallet is becoming a key way of identifying unique grower actions and for comparing grower activity

henriquemoreiraa:
## Description

Wallet filter in the growers tool

**Issue(s) addressed**

- Resolves #952 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

## Other useful information
wallet

### DIFF
--- a/src/models/FilterGrower.js
+++ b/src/models/FilterGrower.js
@@ -61,40 +61,8 @@ export default class Filter {
    * A fn to count the number of current applied filters
    */
   countAppliedFilters() {
-    let numFilters = 0;
-
-    if (this.id) {
-      numFilters += 1;
+    if (this.organizationId) {
+      return Object.keys(this.getWhereObj()).length;
     }
-
-    if (this.personId) {
-      numFilters += 1;
-    }
-
-    if (this.deviceIdentifier) {
-      numFilters += 1;
-    }
-
-    if (this.firstName) {
-      numFilters += 1;
-    }
-
-    if (this.lastName) {
-      numFilters += 1;
-    }
-
-    if (this.email) {
-      numFilters += 1;
-    }
-
-    if (this.phone) {
-      numFilters += 1;
-    }
-
-    if (this.organizationId && this.organizationId !== ALL_ORGANIZATIONS) {
-      numFilters += 1;
-    }
-
-    return numFilters;
   }
 }


src/models/FilterGrower.js
```diff
@@ -16,6 +16,10 @@ export default class Filter {
       where.person_id = this.personId;
     }
 
+    if (this.wallet) {
```
nmcharlton, line 19: Need to add `wallet` to `countAppliedFilters()` below

src/components/FilterTopGrower.js
```diff
@@ -166,6 +169,16 @@ function FilterTopGrower(props) {
                 onChange={(e) => setEmail(e.target.value)}
                 onKeyDown={handleEnterPress}
               />
+              <TextField
```
nmcharlton, line 172: Can we move this one place up to be after Last Name?
That seems like a more logical location.

nmcharlton (changes requested):

src/models/FilterGrower.js
nmcharlton: Feel free to refactor `countAppliedFilters()` so we don't have to do this for each new filter 😄 (but no obligation to do so).

src/models/FilterGrower.js
```diff
@@ -57,40 +61,8 @@ export default class Filter {
...
-    }
-
-    if (this.deviceIdentifier) {
-      numFilters += 1;
-    }
-
-    if (this.firstName) {
-      numFilters += 1;
+    if (this.organizationId) {
+      return Object.keys(this.getWhereObj()).length;
```
nmcharlton, line 65: Note that we need to account for one case where a non-null/undefined (default) value doesn't contribute towards the filter count:
```
this.organizationId === ALL_ORGANIZATIONS
```
See the original implementation:
```
if (this.organizationId && this.organizationId !== ALL_ORGANIZATIONS) {
  numFilters += 1;
}
```

henriquemoreiraa: I've made it some tests and when I don't put the if statement, I receive an undefined value for `organization_id`,  so the `Object.keys.length` catches that. In the `getWhereObj` function there isn't a case for `ALL_ORGANIZATIONS`, only if it is different, so when I apply a filter the `organization_id` disappears then it doesn't contribute for the filter count.

[Screencast from 2022-12-12 13-57-29.webm](https://user-images.githubusercontent.com/98126579/207106654-5be4ed86-e0ec-43c4-844e-2bd39636aa49.webm)

nmcharlton: Thanks @henriquemoreira, that makes sense. Nice work!

nmcharlton (approved):

PR #975, opened 2022-12-07
fix: legacy growers count dash stat not working

### LINKED ISSUE #971: Legacy Growers count dash stat not working
The Growers dash stat card adds `?active=true` to the query.
This is not valid in the new microservice API world and should be omitted.
Screenshot 2022-12-03 at 15 56 18

CODE-Y02:
## Description

I made some changes in DashStatGrowerCount function inside    src\components\DashStat.container.js  .
inside it we are creating growerFilter instance from FilterModel with active == true , but backend api does not accecpt it , 
therefore i have got rid of active value and to be honest we can also get rid of growerFilter instance .

But heres a Problem or More of **API** **issue** i want to address , It make sense to count active growers rather than all growers but current model or api dosen't support it . So we need to find is this really a api problem as we get response from server i assume it is but my knownlege and experiense is nothing so i cant be sure . 
If its api related issue than previous syntex was correct then just undo my changes .

for now everything works fine 

**Issue(s) addressed**
legacy growers count dash stat not working

- Resolves #971

- [x]  Bug fix

## Issue
https://github.com/Greenstand/treetracker-admin-client/issues/971

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/components/DashStat.container.js
+++ b/src/components/DashStat.container.js
@@ -100,9 +100,7 @@ function DashStatVerifiedCaptures(props) {
 }
 
 function DashStatGrowerCount(props) {
-  const growerFilter = new FilterModel({
-    active: true,
-  });
+  const growerFilter = new FilterModel();
 
   const [totalGrowerCount, setTotalGrowerCount] = useState(null);
 


nmcharlton (approved):
Thanks @CODE-Y02!

PR #978, opened 2022-12-11
Fix filtering by organizations using legacy auth and queries to microservices

### LINKED ISSUE #977: Only show and allow filtering by orgs user is permitted to see (using legacy api for auth and new apis for queries)

gwynndp:
## Description

- update SelectOrg to use uuid
- add way to get the org uuid and list of child org uuids from AppContext after login
- update tests to work
- refactor to make query string to use whereNulls, whereNotNulls, whereIns arrays
- adjust filters to work with the new query arrays (tags, organizations, species isn't in filters currently so not tested)

**Issue(s) addressed**

- Resolves #977 
- Relates to api update (https://github.com/Greenstand/treetracker-query-api/pull/268)

- [x] Enhancement
- [x] Bug fix
- [ ] Refactor

### DIFF
--- a/src/api/apiUtils.js
+++ b/src/api/apiUtils.js
@@ -42,10 +42,10 @@ export function getOrganizationUUID() {
 
 export const makeQueryString = (filterObj) => {
   // log.debug('makeQueryString 1 ----->', filterObj);
-  let arr = [];
-  let whereNulls = [];
-  let whereNotNulls = [];
-  let whereIns = [];
+  const arr = [];
+  const whereNulls = [];
+  const whereNotNulls = [];
+  const whereIns = [];
   for (const key in filterObj) {
     if (
       filterObj[key] !== undefined &&


src/context/VerifyContext.js
```diff
@@ -138,6 +144,21 @@ export function VerifyProvider(props) {
     //set loading status
     setIsLoading(true);
 
+    if (!filter.organizationId && orgId === null) {
```
nmcharlton, line 147: This looks like duplicate code (from `GrowerContext`). Given the complexity of the logic, it would definitely be worth making a common function that does this.

src/models/Filter.js
```diff
@@ -231,22 +227,18 @@ export default class Filter {
...
     }
 
-    if (this.tokenId && this.tokenId !== 'All') {
-      numFilters += 1;
-    }
-
-    if (this.status) {
-      numFilters += 1;
-    }
+    // if (this.tokenId && this.tokenId !== 'All') {
```
nmcharlton, line 239: What's the reason for commenting this out?

nmcharlton (changes requested):

src/models/Filter.js
gwynndp: Mental blip.  I was only considering Verify where we aren't currently filtering by token.

src/common/utils.js
```diff
@@ -10,6 +12,34 @@ export const getVerificationStatus = (active, approved) => {
   }
 };
 
+export const setOrganizationFilter = (filter, orgId, orgList) => {
+  // if orgId has a value filter by orgId and sub-orgs, don't include null org ids
+  if (filter.organization_id === ALL_ORGANIZATIONS && orgId) {
```
nmcharlton, line 17: My understanding from the `if` clauses in this function is that only one can be true.
To make that a bit clearer, I'd suggest a chain of `if ... else if ...`

nmcharlton (approved):

Siddhant-K-code (commented):
Possibly unrelated file changes are failing tests. See [here](https://github.com/Greenstand/treetracker-admin-client/actions/runs/3977030788/jobs/6817874725#step:9:2994) 👀
 
image

nmcharlton:
@Siddhant-K-code That's not the test that's failing; it's this one:
```
 FAIL  src/components/tests/verify.test.js (5.315 s)
  Verify
    with default values
      ✓ renders number of applied filters (1725 ms)
      ✕ renders capture details (1241 ms)
      ✓ renders grower details (659 ms)
      ○ skipped renders filter top
      ○ skipped renders captures gallery
      ○ skipped renders edit planter

  ● Verify › with default values › renders capture details

    TypeError: Cannot read properties of undefined (reading 'then')

      70 |
      71 |     try {
    > 72 |       api.getCaptureTags([captureId]).then((tags) => {
         |       ^
      73 |         console.log('All tags: ', tags);
      74 |         setState({ ...state, tags: tags[0] });
      75 |       });
```

src/common/utils.js
gwynndp: It's tricky to come up with variable names.  The filter.organization_id is the filter selected and the "orgId" is the uuid of the logged in user's organization.  

I'll write a comment, but I'm open to better suggestions.

nmcharlton: How about `userOrgId`?

src/api/apiUtils.js
```diff
@@ -39,3 +39,51 @@ export function getOrganizationId() {
 export function getOrganizationUUID() {
   return session.user?.policy?.organization?.uuid || null;
 }
+
+export const makeQueryString = (filterObj) => {
+  // log.debug('makeQueryString 1 ----->', filterObj);
+  const arr = [];
```
nmcharlton, line 45: None of `arr`, `whereNulls`, `whereNotNulls` or `whereIns` should be declared `const`.
They're all potentially modified after declaration.

nmcharlton (changes requested):
Just a minor comment on the use of `const`. Otherwise, looks great!

src/api/apiUtils.js
```diff
@@ -42,10 +42,10 @@ export function getOrganizationUUID() {
 
 export const makeQueryString = (filterObj) => {
   // log.debug('makeQueryString 1 ----->', filterObj);
-  const arr = [];
```
nmcharlton: Sorry @gwynndp, I'm being an idiot. You were right to make these `const`.
I'm still getting up to speed with all this modern JS stuff 🤣 

I'll change it back.

nmcharlton (approved):

PR #994, opened 2023-01-01
fix: update api queries for legacy dash stats

### LINKED ISSUE #993: Legacy dash stats show incorrect values for captures
We're pointing at the legacy API but not filtering by status.
The proper fix for this is to move all users to the new reporting cards (#654), but we're not quite ready for that yet on the API side.
In the meantime, we can just update the queries used for the legacy cards to match those in the Verify and Captures tools.

Screenshot 2023-01-01 at 16 43 21

nmcharlton:
## Description

Interim fix to legacy dash stats cards.
Stats for captures now point at new query API to match numbers in Verify tool.

**Issue(s) addressed**

- Resolves #993 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Capture counts are retrieved from the legacy API and no filters are applied.

Screenshot 2023-01-01 at 16 43 21

**What is the new behavior?**
Captures counts are retrieved from the new query API by status using `/raw-captures?status=unprocessed` for unverified captures and the `/v2/captures` endpoint for approved.
For total captures (which should exclude rejected raw captures) I've combined these two values for consistency.

Screenshot 2023-01-01 at 17 55 40

## Breaking change

No.

## Other useful information
Note that we're not currently doing any filtering by the organization of the logged-in user, but that's consistent with the other parts of the Admin Panel that have been migrated.
This is something that goes wider than these stats and that we need to fix separately (see #977 and #978).

I've reordered the functions in `treeTrackerApi.js` to make the flow and grouping more logical.

### DIFF
--- a/src/components/reportingCards/ReportingCard.hook.js
+++ b/src/components/reportingCards/ReportingCard.hook.js
@@ -40,7 +40,10 @@ export default function useLoadData(
       },
     });
     const { data } = res;
-    log.debug('load reporting data: ', data);
+    if(!data) {
+        log.error('No data found in response');
+    }
+    log.debug('Reporting data: ', data);
 
     let top;
     if (rows !== undefined) {


src/components/reportingCards/ReportingCard.hook.js
```diff
@@ -40,7 +40,7 @@ export default function useLoadData(
       },
     });
     const { data } = res;
-    log.warn('load reporting data: ', data);
+    log.debug('load reporting data: ', data);
```
Siddhant-K-code, line 43: **Suggestion (non-blocking):** We can make it more robust

```suggestion
    if(!data) {
        log.error('No data found in response');
    }
    log.debug('Reporting data: ', data);

```

src/api/treeTrackerApi.js
```diff
@@ -137,35 +137,37 @@ export default {
...
-  getCaptureCount(filter) {
+  /**
+   * Captures
+   */
+  getCaptures({ ...params }) {
     try {
-      const query = `${API_ROOT}/api/${getOrganization()}trees/count?where=${JSON.stringify(
-        filter.getWhereObj()
-      )}`;
+      let filterObj = { limit: 25, offset: 0, ...params };
```
Siddhant-K-code, line 145: **Question:** Any specific reason to limit this to 25? 🤔

nmcharlton: Not sure how this got in there. The API defaults to 25 if no limit is specified, so this is unnecessary.
Now removed.

PR #1010, opened 2023-01-23
Filter bookmark

### LINKED ISSUE #915: Persist filter parameters in URL
## User Story
**As an** Admin Panel user **I want** to bookmark or link directly to pre-filtered views on each of the Admin Panel tools that support filtering **so that** I don't have to set my preferred filters each time I open the tool, and can share direct links to other users.

## Feature Description
We need to encode the current filter selection in the URL as a query string and have it update each time the user changes the filters. When navigating to a valid filter URL, the filter should be applied automatically and the values should populate the filter dialog.

This should be applied as uniformly as possible across all tools with filters:
* Verify
* Captures
* Capture Matching
* Earnings
* Payments
* Growers
* Stakeholders

The mechanism should be built in a way that does not require additional maintenance when new filter parameters are added to any existing filter.

### LINKED ISSUE #1018: Filter settings reset when the filter is collapsed
**Issue:**
Current filter settings reset when the filter is collapsed to better see the results.  So when a user opens the filter again they can't pick up where they left off, instead they have to re-enter everything.

**Applies to:** 
- Verify filter
- Captures filter 
- Growers filter

**Desired behavior:**
- The filter settings should remain while collapsed and be visible as unchanged when the filter is opened again so that the user can clearly see what filters got them to the visible results.

- The filter settings should only be reset when the Reset button is clicked or when the user leaves or refreshes the page.

### LINKED ISSUE #1090: Adding page number in page url (enhancement)
Along with adding the filter values in the url (see issue #915 ), for certain pages such as Verify, should the page number, along with the image size and number of images shown be also added as query parameters?

SuspenseFallback:
## Description
Introduces filter persistence and linkability for the following tools:
* Verify
* Captures
* Growers

When a user applies a filter in one of these tools, the URL search parameters are updated with the filter values and pagination parameters. This provides several benefits:
* Filters and pagination persist on page refresh
* A common filter can be bookmarked in the browser
* Tools can easily link to each other with filtered views, e.g., unverified captures for a specific grower can be linked directly from the Grower card or Grower Detail drawer. No more copying and pasting IDs 😃 
* Changes in filter and pagination are added to the history, so the user can return to a previous filter configuration by going back in the browser

**Issue(s) addressed**

- Resolves #915 
- Resolves #1090
- Resolves #1018

Note that filters for Capture Matching, Earnings, Payments and Stakeholders are not yet supported.
These can be addressed under separate issues using the same pattern and utility functions.

- [x] Enhancement

## Issue

Filters do not persist.

**What is the new behavior?**

Filters stay in the URL.

## Breaking change

No.

## Other useful information
Screenshot 2023-02-26 at 22 14 29
Screenshot 2023-02-26 at 22 17 38
Screenshot 2023-02-26 at 22 18 05


nmcharlton:
Still to do before merging:
- [ ] Add some tests

src/context/GrowerContext.js
```diff
@@ -29,26 +35,65 @@ export const GrowerContext = createContext({
...
     if (orgId !== undefined) {
       load({ signal: abortController.signal });
-      // getCount();
+      // getCount({ signal: abortController.signal });
     }
     return () => abortController.abort();
   }, [filter, pageSize, currentPage, orgId]);
 
+  useEffect(() => {
+    if (!searchParams) {
```
gwynndp, line 83: Could this be part of the previous useEffect, so that there is only one dealing with searchParams?

gwynndp, line 83: Could one useEffect handle what happens whether there are or aren't searchParams?

src/models/Filter.js
```diff
@@ -198,4 +198,27 @@ export default class Filter {
 
     return numFilters;
   }
+
+  toSearchParams() {
```
gwynndp, line 202: Let's add the methods in the same way for consistency.  Filter.toSearchParams() or move the .fromSearchParams() inside the class?

src/context/CapturesContext.js
```diff
@@ -31,19 +39,43 @@ export const CapturesContext = createContext({
...
   const [filter, setFilter] = useState(
-    new FilterModel({ organization_id: ALL_ORGANIZATIONS })
+    FilterModel.fromSearchParams({
+      organization_id: ALL_ORGANIZATIONS,
+      ...filterParams,
+    })
   );
 
   useEffect(() => {
+    handleQuerySearchParams({
```
gwynndp, line 70: Does this need to be inside a conditional like in the Verify and Grower contexts?

gwynndp (dismissed):
Nice solution!

gwynndp:
When testing out the back button with the Capture Filter the filter fields aren't being updated to match the url filters

Test filter data: 
> startDate: 2023-02-17
> endDate: 2023-02-18
> device_identifier: something
> wallet: gweeple@yahoo.com
> organization: test1
> organization_id: ["8b2628b3-733b-4962-943d-95ebea918c9d"]
> grower_account_id: 93a026d2-a511-404f-958c-a0a36892af0f
> tag: Any
> whereNotNulls: ["tag_id"]
> captureUUID: f41f8c5a-c6ad-4c76-a747-46b710e94838
> captureRefId: 999939

Before back button:
https://monosnap.com/file/Jl3tF7djMaCiCz9uLFdCpc4NzrIz83

After back button (the capture reference id and capture uuid fields should be filled): 
https://monosnap.com/file/YuLh586z47GWshCQj4BhY37fA3UdAJ

URL filter string:
 `http://localhost:3001/captures
?uuid=f41f8c5a-c6ad-4c76-a747-46b710e94838
&captureId=999939
&startDate=2023-02-17
&endDate=2023-02-18
&grower_account_id=93a026d2-a511-404f-958c-a0a36892af0f
&device_identifier=something
&wallet=gweeple%40yahoo.com
&species_id=ALL_SPECIES
&organization_id=8b2628b3-733b-4962-943d-95ebea918c9d
&tokenId=All`

src/models/Filter.js
nmcharlton: TIL that JS has a `static` keyword for just this purpose 😄 
I'll move `fromSearchParams()` inside as a static method. Thanks!

src/context/CapturesContext.js
nmcharlton: I think it's the others that are incorrect.
I suspect the `if (searchParams)` check is an artefact of a previous way of doing this.
We probably shouldn't check it as we're not using it within this `useEffect`. I'll do some more testing.

src/context/GrowerContext.js
nmcharlton: See other related comment. I'll remove the `if (searchParams)` check from the other `useEffect`.

PR #1025, opened 2023-02-16
fix: not set org filter

### LINKED ISSUE #1016: Not Set is not working for organization filter when Admin user
When logged in as Admin the queries are the same for "All" and "Not set" in the organization filter

Issue for all filters: Verify, Captures, Growers

https://monosnap.com/file/9Jtj0CaTIfPhKwFRYY4foa4sDdMDwu

henriquemoreiraa:
## Description

The "Not set" option in organization filter now shows different data than the "All" option.

**Issue(s) addressed**

- Resolves #1016 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change


nmcharlton (changes requested):
I don't think this is the correct way to fix this bug. We're still using `'ORGANIZATION_NOT_SET'` as a placeholder string for comparison, but we've removed the shared definition of it. This means we're more susceptible to typo bugs.

Everywhere we reference the string literal `'ORGANIZATION_NOT_SET'`, we should replace it with a variable that's defined once. I notice that we already had the variable `ORGANIZATION_NOT_SET` defined in three places – you've changed one of these, which is likely to cause problems and confusion. Instead, consolidate the three and only change the value to `null` when we populate the API query.

henriquemoreiraa:
> I don't think this is the correct way to fix this bug. We're still using `'ORGANIZATION_NOT_SET'` as a placeholder string for comparison, but we've removed the shared definition of it. This means we're more susceptible to typo bugs.
> 
> Everywhere we reference the string literal `'ORGANIZATION_NOT_SET'`, we should replace it with a variable that's defined once. I notice that we already had the variable `ORGANIZATION_NOT_SET` defined in three places – you've changed one of these, which is likely to cause problems and confusion. Instead, consolidate the three and only change the value to `null` when we populate the API query.

I got a little lost, I was trying to fix the filter input and didn't realize I did that, I couldn't find where the API was calling it at first, so I tried it that way haha!

henriquemoreiraa:
```
    const finalFilter = setOrganizationFilter(
      filter.getWhereObj(), <--
      orgId,
      orgList
    );
```
The first name input field in the growers filter wasn't working, so I removed the `getWhereObj()` from the `filter`, tested it and it started working...

nmcharlton (approved):
Good catch!

PR #1032, opened 2023-02-21
#888 - Grower: new filter by amount of captures uploaded was added. 

### LINKED ISSUE #888: Grower: new filter by amount of captures uploaded.
It would be great to have a simple filter
0
1-25
more than 25

OlhaD:
## Description

#888 - Grower: new filter by amount of captures uploaded was added. 

When I apply the filter, API is called with the parameters below:
https://dev-k8s.treetracker.org/query/v2/growers?capturesAmount_min=0&capturesAmount_max=0&limit=24&offset=0
https://dev-k8s.treetracker.org/query/v2/growers?capturesAmount_min=1&capturesAmount_max=25&limit=24&offset=0
https://dev-k8s.treetracker.org/query/v2/growers?capturesAmount_min=25&limit=24&offset=0

**Issue(s) addressed**

- Resolves #888 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change


src/models/FilterGrower.js
```diff
@@ -55,6 +56,24 @@ export default class Filter {
...
+
+    if (
+      !isNaN(this.capturesAmount_max) ||
+      (this.capturesAmount_range && !isNaN(this.capturesAmount_range.max))
+    ) {
+      where.capturesAmount_max = !isNaN(this.capturesAmount_max)
+        ? this.capturesAmount_max
+        : this.capturesAmount_range.max;
+    }
+
```
OlhaD, line 76: // Problem: this method is called twice.
        //    1st call (from GrowerContext.js - load()): 'this' is an object with 'capturesAmount_range' property. Return object has property 'capturesAmount_min'
       //    2nd call (from growers.js - getGrowers()): 'this' is an object with 'capturesAmount_min' property.
// Workaround: initialize 'where' object with properties which are available
// Better solution: make the only one call

// For the same reason 'firstName' and 'lastName' filters don't work at the moment

nmcharlton: I think you're right – `GrowerContext.js` shouldn't call `getWhereObj()` directly, since this returns the object used the query the API. We're effectively converting the object twice.

Instead, I think `setOrganizationFilter()`should modify a copy of the `FilterGrower` object.

As an aside, we're doing the same thing in `VerifyContext.js` and `CapturesContext.js` – we'll need to apply the same fix there.

Well spotted 😃

nmcharlton: There are a some other filtering bugs that _may_ be related to this: #1016, #1034, #1035
@gwynndp: Can you coordinate with @OlhaD the fix for this works coherently across the different filters?
I'm happy for everything to be fixed under this PR, if it's all caused by the same thing.

OlhaD: @nmcharlton I looked at all 3 issues you mentioned and seems that they are not related to this one.

nmcharlton (approved):
👏

Issue #855, opened 2022-09-30
Inbox: Page fails to load if user handle not found


nmcharlton:
When a user loads the _Inbox_ page, we query their messages on the messaging API.
If the user is not registered for messaging, the API returns 404, but it is not handled correctly in `api/messaging.js` and the page crashes.

AbdelrhmanX7:
@nmcharlton  can i work on this task?

spomberg:
Is this issue still available?

nmcharlton:
@spomberg It is! Let us know if you need any help with it.

spomberg:
@nmcharlton I'm having trouble reproducing this issue, are there any credentials for a user that is not registered for messaging available?

nmcharlton:
@spomberg You can create a new test user through the User Manager page on the Admin Panel. First add a user with the Admin role, then use the key icon against that user to set a password.

pranavkparti:
This issue is fixed, so it should be closed.

PR #1036, opened 2023-02-28
Fix/user handle not found

spomberg:
## Description

Currently, if a user who hasn't been authorized for messaging tries to access the messaging page the API will respond with a 404 error and the page will crash.

The error handling was refactored so that page doesn't crash and the user sees a message on the page asking them to contact technical support to request messaging access.

- Resolves issue #855 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

If a user is not authorized for messaging, the API responds with error 404 causing the page to crash.

**What is the new behavior?**
Instead of crashing the page displays "Your user is not yet configured for messaging access. Please contact technical support." as an error message, as seen below.

image

## Breaking change

None

## Other useful information
Added a new Cypress test to messaging2.spec.py.js and also changed a setting that was preventing tests from running.

### DIFF
--- a/src/context/MessagingContext.js
+++ b/src/context/MessagingContext.js
@@ -185,14 +185,14 @@ export const MessagingProvider = (props) => {
         } else {
           groupMessageByHandle(res.messages);
         }
-      } else {
-        if (res.status === 404 && res.message === 'Author handle not found') {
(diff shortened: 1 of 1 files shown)


nmcharlton (approved):
Thanks @spomberg!

PR #1039, opened 2023-02-28
Edit Growers: Allow user to specify Grower Image URL #845

### LINKED ISSUE #845: Edit Growers: Allow user to specify Grower Image URL
In the Edit Grower dialog, users can select the best image of a grower from a gallery populated from Treetracker app selfies.
Users should be able to specify an image URL so they can use their own hosted images instead.

Note: The design and implementation should be coherent with #192 (image upload).
Once both of these issues are implemented, users will be able to:
1. Select an existing image
2. Upload a new image
3. Use an externally-hosted image via a URL
Screenshot 2022-09-19 at 18 50 00

cvirgili438:
## Description

First all, the image wasn't render before, because props properties doesn't match with the properties declared in the edit panel. so i did modifify them.
Before 
img not found
After 
img not found
after that, i made some changes for the arrays of nulls, i denied them in the images, because they render a white square, and that seems a little bad.

P.S: i cannot try if the save is working. When i try to click in the button save, thow me a error 
img not found
Because  i started  in this project a few days ago, i dont know much about that.
did i touch some var to let it happen? 
Thanks for the patience 

**Issue(s) addressed**

- Resolves #845 

- [ x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

**What is the new behavior?**

## Breaking change

### DIFF
--- a/src/api/growers.js
+++ b/src/api/growers.js
@@ -90,14 +90,16 @@ export default {
       }
       const { id } = growerUpdate;
       const growerQuery = `${TREETRACKER_API}/grower_accounts/${id}`;
-      delete growerUpdate.id;
+
+      let newBody = Object.assign({}, growerUpdate);
+      delete newBody.id;
       return fetch(growerQuery, {
         method: 'PATCH',
         headers: {
           'content-type': 'application/json',
           Authorization: session.token,
         },
-        body: JSON.stringify(growerUpdate),
+        body: JSON.stringify(newBody),
       }).then(handleResponse);
     } catch (error) {
       handleError(error);


src/components/EditGrower.js
```diff
@@ -126,6 +141,16 @@ const EditGrower = (props) => {
             }
             onSelectChange={handleChange}
           />
+          <TextField
+            className={classes.textInput}
+            label="Image Custom URl"
```
nmcharlton, line 146: Should be `URL` (all caps)

src/components/EditGrower.js
```diff
@@ -126,6 +141,16 @@ const EditGrower = (props) => {
...
+            error={!isImgURL(customUrl)}
```
nmcharlton, line 147: This shows as an error when the user opens the dialog before they've done anything – not a great user experience.
Instead, only check when they blur the input, and display a note explaining the error ("Invalid image URL").

src/components/EditGrower.js
```diff
@@ -38,25 +38,40 @@ const EditGrower = (props) => {
...
+        const notAllowNull =
+          selfies !== undefined && selfies.includes(null)
+            ? selfies.filter((e) => e !== null)
+            : selfies;
         setGrowerImages([
-          ...(grower.imageUrl ? [grower.imageUrl] : []),
-          ...(selfies || []).filter((img) => img !== grower.imageUrl),
+          ...(grower.image_url ? [grower.image_url] : []),
+          ...(notAllowNull || []).filter((img) => img != grower.image_url),
+          customUrl === '' ? null : customUrl,
```
nmcharlton, line 56: I like this way of doing it, but we should put `customUrl` at the front of the array.

src/components/EditGrower.js
```diff
@@ -126,6 +141,16 @@ const EditGrower = (props) => {
             }
             onSelectChange={handleChange}
           />
+          <TextField
+            className={classes.textInput}
+            label="Image Custom URl"
+            error={!isImgURL(customUrl)}
+            onChange={(e) => {
+              if (isImgURL(e.target.value)) {
+                setCustomUrl(e.target.value);
```
nmcharlton, line 150: We should auto-select this image when the user enters a valid URL (by calling `handleChange('image_url', e.target.value)`. It's fair to assume that if they're entering a URL, that's the image they want to use.

nmcharlton (changes requested):
This is looking good. Just a few comments around duplication and user experience.

src/components/EditGrower.js
```diff
@@ -38,25 +38,40 @@ const EditGrower = (props) => {
...
+  const [customUrl, setCustomUrl] = useState('nothing');
 
   useEffect(() => {
     async function loadGrowerImages() {
       if (grower?.id) {
         setLoadingGrowerImages(true);
         const selfies = await growerContext.getGrowerSelfies(grower.id);
         setLoadingGrowerImages(false);
-
+        const notAllowNull =
```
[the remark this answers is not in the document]
cvirgili438: i read requested changes and i did this.
i delete that const notAllowNull,  and apply the filter in the component "ImageScroller" Instead.
image

And i worked with customImageURL as you told me 
image
with nulls, is more easy to work in error's handle

src/components/EditGrower.js
```diff
@@ -38,25 +38,40 @@ const EditGrower = (props) => {
...
-  }, [grower]);
-
+  }, [grower, customUrl]);
+  function isImgURL(url) {
+    if (typeof url === 'object') {
+      return url === null
+        ? false
+        : /\.(gif|jpe?g|tiff?|png|webp|bmp)$/i.test(url.image_url);
+    }
+    if (url === null || url === undefined) {
```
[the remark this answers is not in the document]
cvirgili438: I change that function, i delete that if(), because typeof null is an object too. So i was repeating code for no reason
image

src/components/EditGrower.js
cvirgili438: i solved thise with a state
image
image
image

src/components/EditGrower.js
cvirgili438: i has some issues about that.
i modify ImageScroller where the properties was imageUrl instead image_url, that works in half, in the first render, the main image was autoselected, but when i introduce a valid Custom URL, i can't making the new img into a selected one's, i don't know why.
Tomorrow with more time, i'll try to search a solution

src/components/EditGrower.js
```diff
@@ -38,25 +38,34 @@ const EditGrower = (props) => {
...
+          ...(selfies || []).filter((img) => img != grower.image_url),
         ]);
       }
     }
-
     setGrowerUpdate(null);
     loadGrowerImages();
-  }, [grower]);
-
+  }, [grower, customImageUrl]);
```
nmcharlton, line 62: Including `customImageUrl` here as a trigger here causes us to reload images and reset the dialog, meaning the custom image is not auto-selected.

Instead, just add the custom URL to the list of images when you pass it to `ImageScroller`:
```js
          <ImageScroller
            images={[
              customImageUrl === '' ? null : customImageUrl,
              ...growerImages
            ].filter((e) => e !== null)}
```

src/components/EditGrower.js
```diff
@@ -117,15 +142,28 @@ const EditGrower = (props) => {
       <DialogContent>
         <Grid container direction="column" className={classes.container}>
           <ImageScroller
-            images={growerImages}
-            selectedImage={growerUpdate?.imageUrl || grower.imageUrl}
+            images={growerImages.filter((e) => e !== null)}
```
nmcharlton, line 145:
```js
            images={[
              customImageUrl === '' ? null : customImageUrl,
              ...growerImages
            ].filter((e) => e !== null)}
```

src/components/EditGrower.js
```diff
@@ -38,25 +38,34 @@ const EditGrower = (props) => {
...
-          ...(selfies || []).filter((img) => img !== grower.imageUrl),
+          customImageUrl === '' ? null : customImageUrl,
+          ...(grower.image_url ? [grower.image_url] : []),
+          ...(selfies || []).filter((img) => img != grower.image_url),
         ]);
       }
     }
-
     setGrowerUpdate(null);
     loadGrowerImages();
```
nmcharlton, line 61: Call `setCustomImageUrl(null)` here to make sure we properly reset the dialog when the selected grower changes.

nmcharlton (changes requested):

nmcharlton:
@cvirgili438 I've just noticed that we're using the wrong API in `api/growers.js updateGrower()`.
It should be 
```js
const TREETRACKER_API = process.env.REACT_APP_TREETRACKER_API_ROOT;
...
const growerQuery = `${TREETRACKER_API}/grower_accounts/${id}`;
```
Can you make the change in this PR? It should make testing possible.
Sorry for the confusion.

cvirgili438:
> @cvirgili438 I've just noticed that we're using the wrong API in `updateGrower()` (in `api/growers.js`). It should be
> 
> ```js
> const TREETRACKER_API = process.env.REACT_APP_TREETRACKER_API_ROOT;
> ...
> const growerQuery = `${TREETRACKER_API}/grower_accounts/${id}`;
> ```
> 
> Can you make the change in this PR? It should make testing possible. Sorry for the confusion.

I already changed in my commit
```
const TREETRACKER_API = process.env.REACT_APP_TREETRACKER_API_ROOT;
...
const growerQuery = `${TREETRACKER_API}/grower_accounts/${id}`;
```

But, seems to appear a new error around the post 
image

nmcharlton:
Looks like you're uncovering a few other bugs!
Just like you had to change `imageUrl` to `image_url`, can you change `firstName` to `first_name` and `lastName` to `last_name` in `EditGrower.js`?

You'll also need to remove `id` from `growerUpdate` when you pass it to API query.

cvirgili438:
> Looks like you're uncovering a few other bugs! Just like you had to change `imageUrl` to `image_url`, can you change `firstName` to `first_name` and `lastName` to `last_name` in `EditGrower.js`?
i already change that, and seems work,
> 
> You'll also need to remove `id` from `growerUpdate` when you pass it to API query.
i just remove `id`  from `growerUpdate`  in `EditGrower.js` 
And this is what appear when y try to save the changes 
image

Also i tried to change id in updateGrower to `grower_account_id` but still not working 
image
 
 Another question, i need to change organizationId for organization_id too ? 
 
image

I saw the PR #1025 and he use organization_id,  should  i change it ?  correct me if  i'm wrong.

src/components/EditGrower.js
```diff
@@ -38,40 +38,67 @@ const EditGrower = (props) => {
...
+    }
+    return regex.test(url);
+  }
   async function handleSave() {
     if (growerUpdate) {
       setSaveInProgress(true);
       // TODO handle errors
       await growerContext.updateGrower({
-        id: grower.id,
+        // id: grower.id,
```
nmcharlton, line 75: You still need to pass the `id` so that `api/growers.updateGrower()` knows where to route the request.
Just remove the `id` before populating the body of the request in `api/growers.updateGrower()`.

nmcharlton (changes requested):

nmcharlton:
> I saw the PR https://github.com/Greenstand/treetracker-admin-client/pull/1025 and he use organization_id, should i change it ? correct me if i'm wrong.

You are correct 👍

src/components/EditGrower.js
cvirgili438: Got it, 
I just delete Id property before the fetch and it work 
image
The changes which i do in  one user is working but, i have a little problem.
when i click the button save, did the change and keeping in a loop for an uncaugth promise, 
image
It's like a the state SaveInProgress never set in false again, and keeping in true

nmcharlton: That's because you're deleting the `id` from an object that is used afterwards by the calling function:
```
  const updateGrower = async (payload) => {
    if (payload.organizationId === 'ORGANIZATION_NOT_SET') {
      payload.organizationId = null;
    }
    await api.updateGrower(payload);
    const updatedGrower = await api.getGrower(payload.id);  // <-- This expects id to still be there
```

You shouldn't modify the object that's passed to the function. Instead, create a new object without `id`:
```
      const { id, ...payload } = growerUpdate;
      const growerQuery = `${TREETRACKER_API}/grower_accounts/${id}`;
      return fetch(growerQuery, {
        method: 'PATCH',
        headers: {
          'content-type': 'application/json',
          Authorization: session.token,
        },
        body: JSON.stringify(payload),
      }).then(handleResponse);
```

cvirgili438: Thanks, i forgot that function. 
Already done the changes

nmcharlton: Have you pushed your latest changes? I don't see them.

cvirgili438: I'm sure  I did the push, in 2 hour I'm going to my house an I'll check my computer.

cvirgili438: > Have you pushed your latest changes? I don't see them.


I already pushed, sorry that night i did the commit but seems i forgot to push to this branch 
image
image

nmcharlton (approved):
Thanks @cvirgili438, great work!

PR #1041, opened 2023-03-04
Feature/organization logo

### LINKED ISSUE #285: Need for organization logo in Admin Panel
We would need to be able to show the logo of the organization. If it is available in the entity table it has to be shown.

spomberg:
## Description

Displays the organization logo instead of Greenstand's if the user belongs to an organization with a logo URL available.

**Issue(s) addressed**

- Resolves #285 

- [x] Enhancement
- [ ] Bug fix
- [ ] Refactor

## Issue

Currently, Greenstand's logo is displayed on the navbar and on the menu no matter what.

**What is the new behavior?**
Implemented a logoPath state that serves as the image source in the IconLogo component, once the user logs in, a request is made to retrieve the organization's logo URL, if there is one available the state will be set to that path so the organization logo is displayed on the navbar and menu instead of Greenstand's. If the user is logged out the logoPath state defaults to  Greenstand's logo.

image

image

## Breaking change

No

### DIFF
--- a/cypress/integration/IconLogo.spec.py.js
+++ b/cypress/integration/IconLogo.spec.py.js
@@ -1,14 +1,8 @@
 describe('Icon Logo', () => {
-  Cypress.on('uncaught:exception', (err, runnable) => {
-    // returning false here prevents Cypress from
-    // failing the tests due to uncaught errors
-    return false;
-  });
-
   it('Displays the organization logo when the user belongs to an organization', () => {
     cy.visit('http://localhost:3001/login');
-    cy.get('#userName').type('test1');
-    cy.get('#password').type('EoCAyCPpW0');
+    cy.get('#userName').type(Cypress.env('test_username'));
+    cy.get('#password').type(Cypress.env('test_password'));
     cy.contains(/log/i).click();
     cy.contains(/earnings/i).click();
     cy.wait(100);
@@ -16,8 +10,8 @@ describe('Icon Logo', () => {
   });
   it('Displays the Greenstand logo when the user does not belongs to an organization', () => {
     cy.visit('http://localhost:3001/login');
-    cy.get('#userName').type('admin');
-    cy.get('#password').type('8pzPdcZAG6&Q');
+    cy.get('#userName').type(Cypress.env('admin_username'));
+    cy.get('#password').type(Cypress.env('admin_password'));
     cy.contains(/log/i).click();
     cy.contains(/earnings/i).click();
     cy.wait(100);
--- a/cypress/integration/IconLogo.spec.py.js
+++ b/cypress/integration/IconLogo.spec.py.js
@@ -1,0 +1,26 @@
+describe('Icon Logo', () => {
+  Cypress.on('uncaught:exception', (err, runnable) => {
+    // returning false here prevents Cypress from
+    // failing the tests due to uncaught errors
+    return false;
+  });
+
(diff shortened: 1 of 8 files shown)


src/components/Login.js
```diff
@@ -142,7 +142,25 @@ const Login = (props) => {
           if (res.status === 200) {
             const token = res.data.token;
             const user = res.data.user;
-            appContext.login(user, token, isRemember);
+            // GET logo URL from API if user belongs to an organization
+            // and apply it to logoPath state before completing login
+            if (user.policy.organization) {
+              const STAKEHOLDER_API =
```
nmcharlton, line 148: Can you move this `getStakeholder(id)` query to `api/stakeholders.js` to avoid duplication?

nmcharlton (changes requested):
Great work @spomberg, just one suggestion.

@sebastiangaertner Is this what you had in mind?
Do we enforce any kind of size restrictions on org logos?

src/components/Login.js
spomberg: I created a `getStakeholder(id)` query in `api/stakeholders.js` which is now being called whenever we need to retrieve the logo URL.
Also, instead of using the IconLogo component on `Login.js` I applied the `logo.svg` directly since we only expect to see the Greenstand logo on that page and it was causing one of the tests to fail.

cypress/integration/IconLogo.spec.py.js
```diff
@@ -0,0 +1,26 @@
...
+    cy.get('#password').type('EoCAyCPpW0');
+    cy.contains(/log/i).click();
+    cy.contains(/earnings/i).click();
+    cy.wait(100);
+    cy.get('img').should('have.attr', 'alt', 'organization logo');
+  });
+  it('Displays the Greenstand logo when the user does not belongs to an organization', () => {
+    cy.visit('http://localhost:3001/login');
+    cy.get('#userName').type('admin');
+    cy.get('#password').type('8pzPdcZAG6&Q');
```
nmcharlton, line 20: We need to find another way to run this test. We can't commit credentials to the repository like this.
Can we somehow prompt for the credentials when running the test or take them from a local config file that's not committed?

nmcharlton (changes requested):

cypress/integration/IconLogo.spec.py.js
spomberg: I do apologize! I came across an older committed integration test that contained credentials and thought this was a common practice, I should have known better. 

I removed all credentials I found on Cypress tests and created a `cypress.env.json` file to store credentials to use on tests, this file has also been added `.gitignore`.

nmcharlton: No worries!

sebastiangaertner:
@nmcharlton can we make it svg mandatory then the size is flexible or would it just read the stakeholder logo url? 
Dimensions are not the issue here or? either square or landscape.

nmcharlton:
@sebastiangaertner How do org logos get into the system? Do orgs send them to us based on some requirement?

The issue is that if we're directly replacing a wide, rectangular Greenstand logo with a square organization logo, it doesn't really fit in the space in the same way.
Judging by the web map, it would seem that all org logos are assumed to be square. If this is the case, we should tweak the design of the org logo in the Admin Panel to be optimised for a square logo.

sebastiangaertner:
@dadiorchen could this be part of Keycloak?

Issue #1072, opened 2023-04-02
Species not working in Captures or Capture Details


nmcharlton:
Related to #1017

This is a quick fix to get species information back into the Captures page, filter and Capture Detail dialog.

Not working:
* Species filter in _Captures_ tool
* Species column in _Captures_ tool
* Species field in Capture Detail dialog

The new Query API expects a UUID for the species, available as `species.uuid` in the legacy Species data.

In the future, species information will be part of a dedicated schema, indexed by UUID.

PR #1076, opened 2023-04-18
fix: switch speciesLookup from using id to uuid

gwynndp:
## Description

Fix display and filtering for species on Captures tool

**Issue(s) addressed**

- Resolves #[1072](https://github.com/Greenstand/treetracker-admin-client/issues/1072)

**Bug Found**
- Reveals a bug in the api that returns duplicate captures when filtering by species, maybe also with other filters

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

### DIFF
--- a/src/api/treeTrackerApi.js
+++ b/src/api/treeTrackerApi.js
@@ -295,7 +295,7 @@ export default {
         .then(handleResponse)
         .then((res) => {
           // log.debug(res);
-          return res[0];
+          return res?.length ? res[0] : null;
         });
     } catch (error) {
       handleError(error);


src/components/CaptureDetailDialog.js
```diff
@@ -323,10 +323,10 @@ function CaptureDetailDialog({
         <Grid item className={classes.box}>
           <Typography className={classes.subtitle}>Tags</Typography>
           <Typography variant="subtitle1">Species</Typography>
-          {species && species.name ? (
+          {species && species[0]?.name ? (
```
nmcharlton, line 326: Has this been consciously changed from an object to an array as part of the migration?
If not, is there an error somewhere? We don't allow multiple species to be assigned to a single capture.

gwynndp: I don't know, sorry I didn't explore it but I'll check if it's coming from the api or somewhere in the client code.

gwynndp: The array result is coming from the API.  It is an array now because we are using the /species endpoint with a filter for the uuid instead of making a request to /species/:id

I don't think it is worth updating the legacy API at this point.  We can make the adjustment to the client when we get the new API.

nmcharlton: Let's hide that workaround in the `getSpeciesById()` method, so that callers still receive a single `species` as a result.
It's fair to assume we'll have 0 or 1 items in the response from the API.

src/context/CaptureDetailContext.js
```diff
@@ -83,7 +83,7 @@ export function CaptureDetailProvider(props) {
     }
 
     return api.getSpeciesById(speciesId).then((species) => {
-      setState({ ...state, species });
+      setState({ ...state, species: species[0] });
```
nmcharlton, line 86: I'd suggest doing the extraction within `api.getSpeciesById()` so that all callers get the species object or `null` (if `species.length === 0`) without having to extract it.

nmcharlton (approved):

PR #1082, opened 2023-04-22
Reject capture

### LINKED ISSUE #1046: Verify: Cannot reject a capture
## Steps to Reproduce
1. Open _Verify_
2. Select a capture
3. Select _REJECT_ from the side drawer
4. Submit

### Current Behaviour
Rejection fails due to an incorrect URL root in the API call: `https://dev-admin.treetracker.org/undefined/raw-captures/:id/reject`

### Desired Behaviour
Rejection succeeds with a call to `https://dev-k8s.treetracker.org/field-data/raw-captures/:id/reject`
Note that the `field-data` URL is defined as the env var `REACT_APP_FIELD_DATA_API_ROOT`.

SuspenseFallback:
## Description

Captures can now be rejected.

**Issue(s) addressed**

- Resolves #1046 

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Rejecting a capture gives `ERROR_404` due to `.env` variable mismatch.

**What is the new behavior?**

Rejecting a capture successfully happens.

## Breaking change

No.

### DIFF
--- a/src/api/treeTrackerApi.js
+++ b/src/api/treeTrackerApi.js
@@ -1,16 +1,17 @@
 import {
-  handleResponse,
-  handleError,
   getOrganization,
+  handleError,
+  handleResponse,
   makeQueryString,
 } from './apiUtils';
+
 import { getVerificationStatus } from '../common/utils';
-import { session } from '../models/auth';
 import log from 'loglevel';
+import { session } from '../models/auth';
 
 // Set API as a variable
 const API_ROOT = process.env.REACT_APP_API_ROOT;
-const FIELD_DATA_API = process.env.REACT_APP_FIELD_DATA_ROOT;
+const FIELD_DATA_API = process.env.REACT_APP_FIELD_DATA_API_ROOT;
 const QUERY_API = process.env.REACT_APP_QUERY_API_ROOT;
 const TREETRACKER_API = process.env.REACT_APP_TREETRACKER_API_ROOT;
 const STATUS_STATES = {
@@ -120,6 +121,7 @@ export default {
   async rejectCaptureImage(capture, rejectionReason) {
     try {
       log.debug('reject capture', capture.id, rejectionReason);
+      console.log('env url', FIELD_DATA_API);
       const query = `${FIELD_DATA_API}/raw-captures/${capture.id}/reject`;
       const data = await fetch(query, {
         method: 'PATCH',
--- a/src/components/SidePanel.js
+++ b/src/components/SidePanel.js
@@ -1,24 +1,25 @@
-import React, { useState, useContext } from 'react';
-import { makeStyles } from '@material-ui/core/styles';
 import {
   Box,
-  Divider,
-  Typography,
-  FormControlLabel,
-  Drawer,
-  RadioGroup,
-  Radio,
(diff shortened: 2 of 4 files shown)


SuspenseFallback:
Due to git errors, I have to close this pull request. I will redo this.

SuspenseFallback:
Sorry for the mistake. I was trying to close another pull request

.env
```diff
@@ -1,2 +1,3 @@
+REACT_APP_WEBMAP_DOMAIN=https://map.treetracker.org
```
nmcharlton, line 1: Looks like you've created this branch from another dev branch rather than `master` and therefore have carried across some commits that aren't related to this fix.

src/api/treeTrackerApi.js
```diff
@@ -120,6 +121,7 @@ export default {
   async rejectCaptureImage(capture, rejectionReason) {
     try {
       log.debug('reject capture', capture.id, rejectionReason);
+      console.log('env url', FIELD_DATA_API);
```
nmcharlton, line 124: Please don't leave `console.log()` statements in the code. If you believe they are useful for debug, use `log.debug()` instead, as it is switched off by default in production but can be enabled if need be.

src/components/GrowerDetail.js
```diff
@@ -193,6 +193,7 @@ const GrowerDetail = ({ open, growerId, onClose }) => {
           setErrorMessage(match.message);
         }
 
+        console.log('grower', match);
```
nmcharlton, line 196: Same comment regarding `console.log()` – use `log.debug()` instead.

nmcharlton: Also, I'm not sure what this logging has to do with the fix.

src/components/SidePanel.js
```diff
@@ -1,24 +1,25 @@
-import React, { useState, useContext } from 'react';
-import { makeStyles } from '@material-ui/core/styles';
 import {
   Box,
+  Button,
```
nmcharlton, line 3: I don't mind you committing linter changes like this if there are functional changes to a file, but in this case it's just the ordering of `import` statements that has changed. That tends to cause merge headaches for other developers.

src/components/Verify.js
```diff
@@ -1,42 +1,42 @@
-import React, { useEffect, useState, useContext, useRef } from 'react';
```
nmcharlton, line 1: Same comment about committing linter-only changes for a file as above.

src/context/VerifyContext.js
```diff
@@ -1,9 +1,11 @@
-import React, { useContext, useState, useEffect, createContext } from 'react';
```
nmcharlton, line 1: More unnecessary changes.

nmcharlton (changes requested):
Only one line in one file has actually been changed to fix this bug, so there shouldn't be 9 files in the PR.

.env
SuspenseFallback: Oh that was why I had so many git errors, thank you!

SuspenseFallback:
I am going to close this as I made multiple mistakes in git, I am trying to redo this.

PR #1101, opened 2023-07-04
Reporting fix

### LINKED ISSUE #1100: Specific reporting cards display empty entries
Some of the reporting cards show empty entries, as seen in the images:

image
image

### LINKED ISSUE #1102: Grower Gender Counts reporting card displays entry incorrectly
The reporting card, when the detailed modal is opened by clicking See More and is closed, the entry disappears (empty value). 
Recording 2023-07-03 at 22 59 20

pranavkparti:
## Description
The Growers, Captures, Unverified Captures, and Average Captures per Grower reporting cards show  'No organization' instead of an empty entry. 

**Issue(s) addressed**
- Resolves #1100 
- Resolves #1102 

- [] Enhancement
- [X] Bug fix
- [ ] Refactor

## Issue

 Some of the reporting cards show empty entries instead of some non-null value.

**What is the new behavior?**
image
image
The reporting cards show 'No organization' instead of an empty entry.

## Breaking change

No.

## Other useful information
None.

### DIFF
--- a/src/components/Home/Home.js
+++ b/src/components/Home/Home.js
@@ -32,6 +32,7 @@ import ReportingCard5 from '../reportingCards/ReportingCard5';
 import ReportingCard6 from '../reportingCards/ReportingCard6';
 import ReportingCard7 from '../reportingCards/ReportingCard7';
 import ReportingCard8 from '../reportingCards/ReportingCard8';
+import ReportingCard9 from '../reportingCards/ReportingCard9';
 import MenuItem from '@material-ui/core/MenuItem';
 import MenuMui from '@material-ui/core/Menu';
 import { format, subDays, formatDistanceToNow } from 'date-fns';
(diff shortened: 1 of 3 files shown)


src/components/reportingCards/ReportingCard1.js
```diff
@@ -15,6 +15,25 @@ export default function ReportingCard1(props) {
     rows
   );
 
+  const formatData = (data) => {
```
nmcharlton, line 18: Rather than duplicating this function across reporting cards, I'd suggest creating a shared utility function that replaces blank entries with "No organization" and giving it a name that indicates that modification.

To be more general, it could take the string you want to use in place of blank entries as an input.

nmcharlton (changes requested):

src/components/reportingCards/ReportingCard1.js
pranavkparti: That was what I originally went with, but there could be other reporting cards later as well. There is a card with gender, and species. But I see you point. We can change the reporting card itself to accept a default value that is displayed if the entry is empty.

pranavkparti: @nmcharlton For now I've just shifted the formatData function to utilities. It would be better to refactor the reporting cards in a separate issue.

Issue #886, opened 2022-10-10
extend filters by showing related organizations or not


sebastiangaertner:
would it be possible to have a default (show all sub orgs) inline select button like this ? Is that even possible? 
![Screenshot 2022-10-10 at 11 34 42](https://user-images.githubusercontent.com/3283800/194832338-6629fd49-0798-40e7-85d7-735de91abf9f.png)

nmcharlton:
I'd suggest we make it a separate toggle that's only enabled when an org is selected in the filter. A toggle inside a dropdown list like this is a little odd.

nmcharlton:
I'm not sure how org filtering is implemented in the new APIs, but the legacy API automatically includes sub-orgs.

@gwynndp @Kpoke How do you see this working?

tosinolarinde:
Hi team. I would like to request access to this issue and be assigned to it so I can begin working on it.

cynthiabondi:
31.03.26 update: @tosinolarinde sent a draft to Karthik yesterday, she was able to work with a picture of the interface she found on the admin page but she has also asked to see the full UI and user flow currently in use

sebastiangaertner:
@tosinolarinde let me know how it is going.  See you on slack.

cynthiabondi:
prototype completed and waiting for review @tosinolarinde

cynthiabondi:
Prototype confirmed by Seb. https://www.figma.com/design/1vVataT9HKeZSH9EU8G7H6/Greenstand?node-id=0-1&p=f The design shows two states:

Toggle off — only the org filter shows, no sub-org dropdown
Toggle on — a "Sub-organizations" dropdown appears (defaulting to "All sub-orgs")

This answers the outstanding default question: include sub-orgs is opt-in (toggle off by default), with "All sub-orgs" as the default selection once enabled.

PR #1124, opened 2023-10-28
Showing Sub-organizations as dropdown from the organization of a Grower

samwel141:
## Description
managed to show the dropdown list of sub-organizations associated with an organization of a Grower

**Issue(s) addressed**
Issue number 886 under progress to be resolved

- [x ] Enhancement
- [ ] Bug fix
- [x] Refactor

- [x ] The commit message follows our guidelines

## Issue

**What is the new behavior?**

now on checking an inline radio button the list of sub-organizations drops from organization name

## Breaking change

May close issue: [#886](https://github.com/Greenstand/treetracker-admin-client/issues/886

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
     "@cypress/react": "^6.0.0",
+    "@cypress/webpack-dev-server": "^2.0.0",
     "@storybook/addon-actions": "^6.5.9",
     "@storybook/addon-links": "^6.5.9",
     "@storybook/addons": "^6.5.9",
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,13 +59,13 @@
         "stream": "npm:stream-browserify",
         "typeface-roboto": "*",
         "uuid": "^9.0.1",
-        "webpack": "^5.89.0",
         "zxcvbn": "^4.4.2"
       },
       "devDependencies": {
         "@babel/preset-react": "^7.18.6",
         "@commitlint/cli": "^11.0.0",
         "@commitlint/config-conventional": "^11.0.0",
+        "@cypress/react": "^6.0.0",
         "@cypress/webpack-dev-server": "^2.0.0",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "stream": "npm:stream-browserify",
     "typeface-roboto": "*",
     "uuid": "^9.0.1",
+    "webpack": "^5.89.0",
```
gwynndp, line 57: Was this intended?  It shouldn't be necessary, but if you ran into a problem then we can create a separate issue for it.

gwynndp:
@samwel141 

Off to a nice start!  Can you link to the issue that this PR resolves?  
I think there are some open questions to answer before we can fully resolve this.

- Do we want to show all the Sub-Orgs for the Organization on each grower, or only show the Organization and Sub-Orgs that they belong to
- For a better UX we may need a way to tell the difference between the Org/Sub-Orgs the grower belongs to and the rest, if we want to list them all there

In the meantime, there is a display problem with your current implementation. It is also displaying on the Grower Cards in a jumbled way.  See below:

Screenshot 2023-11-24 at 12 45 11 PM

On the card we might just want to list all of the organizations as comma-separated values to keep it compact, rather than have a dynamic dropdown.

gwynndp (changes requested):

samwel141:
Hello, thank you so much for your consideration.
I was trying to solve this issue
https://github.com/Greenstand/treetracker-admin-client/issues/886 and my
concern is to know from you if that is what you wanted.
About the change fracturing the UI in cards, I'll work on it but it depends
on your reply to the above question.
Any change in the package.json file was not intended and will rectify it.

Regards,

samwel141:
I have put everything clear in the UI can you give feedback on that also

gwynndp:
Looks much better.  If you remove your package.json and package-lock.json changes then I'll merge.

samwel141:
Removed changes in the package.json and package-lock.json, so it is now clear.

gwynndp (approved):

PR #1201, opened 2026-06-01
fix: show device identifier in grower detail when registration is empty

### LINKED ISSUE #1194: Device identifier is not served by the registration info (field data)
See below. The device identifier only works of IOS not for Android
Image

marsof02:
## Description

Grower detail only loaded device identifiers from `planter-registration`. For many growers (especially when registration rows are missing or lack `device_identifier`), the identifier still exists on captures (`trees.deviceIdentifier`), so the UI showed `---` even though the API had data.

This PR:

- Keeps registration as the primary source and reads alternate field shapes (`device_identifier`, `deviceIdentifier`, nested `field_data` / `fieldData`).
- Falls back to deduped device identifiers from the grower’s most recent captures (up to 100) when registration returns none.
- Fixes grower switching when the previous grower had an empty registration array (reload did not run).
- Handles registration fetch errors without leaving stale device-id state.

**Admin-only workaround** — does not fix missing `device_identifier` on `planter-registration` in field-data / bulk-pack ingestion (follow-up on backend).

**Issue(s) addressed**

- Resolves #1194

- [ ] Enhancement
- [x] Bug fix
- [ ] Refactor

## Issue

Grower detail shows device identifiers only from `GET .../planter-registration`. If that list is empty or rows lack `device_identifier`, the Device Identifier section shows `---`, even when the grower’s captures include `deviceIdentifier` (e.g. dev growers **22711**, **22707**). Switching between growers after viewing one with empty registrations could also leave device IDs stale.

**What is the new behavior?**

Device identifiers are built from registrations when present; otherwise from recent captures. Opening a different grower reloads registration (and capture fallback when needed) correctly. OS label uses manufacturer when available, with shape-based inference for capture fallback.

## Breaking change

No.

## Other useful information

**Test plan (dev-k8s)**

- [ ] Grower **22711** — empty `planter-registration`, captures have `deviceIdentifier` → identifier visible in Grower detail.
- [ ] Grower **22707** — same pattern.
- [ ] Grower with registration device ids (e.g. **22716**) — no regression.
- [ ] Open grower A, then grower B — device ids update for B.

**Follow-up (out of scope)**

- Persist/serve `device_identifier` on `planter-registration` (bulk-pack / field-data pipeline).
- Re-test reporter grower **20692** on prod (not on dev-k8s).
- Optional: Growers list filter applies `deviceIdentifier: null` when Device ID filter is empty.

Manual testing on dev-k8s; no new unit tests added.

### DIFF
--- a/src/components/GrowerDetail.js
+++ b/src/components/GrowerDetail.js
@@ -133,6 +133,91 @@ const useStyle = makeStyles((theme) => ({
   },
 }));
 
+const CAPTURE_DEVICE_LOOKUP_LIMIT = 100;
+
+const isPresentDeviceIdentifier = (value) => {
+  if (value == null) return false;
+  return String(value).trim().length > 0;
+};
+
+const getRegistrationDeviceIdentifier = (reg) => {
+  if (!reg) return undefined;
+  const candidates = [
+    reg.device_identifier,
+    reg.deviceIdentifier,
+    reg.field_data?.device_identifier,
+    reg.field_data?.deviceIdentifier,
+    reg.fieldData?.device_identifier,
+    reg.fieldData?.deviceIdentifier,
+  ];
+  return candidates.find(isPresentDeviceIdentifier);
+};
+
+const getRegistrationManufacturer = (reg) => {
+  if (!reg) return undefined;
+  return (
+    reg.manufacturer ??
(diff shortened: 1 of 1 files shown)


cynthiabondi:
@dadiorchen , can you please review this PR

